PR #21, opened 2023-04-03
updated file naming conventions along with its dependencies

Ishankoradia:


fatchat (approved):
ok but i will 100% get this wrong

PR #31, opened 2023-04-05
29 consistent api and api error messages

Ishankoradia:


fatchat (changes requested):
Put in doc strings for new functions please?

Ishankoradia:
> Put in doc strings for new functions please?

Done

fatchat (approved):

Issue #22, opened 2023-04-03
CRUD-ify the AirbyteService


fatchat:
The AirbyteService offers creation and retrieval of sources, destinations and connections. 

Implement Delete and Update

PR #38, opened 2023-04-10
add update api for source and destination

Abhishek-N:
Fix: #22


fatchat (changes requested):
nice, but we no longer have the AuthBearer middleware, please pull latest changes from main and use auth.CanManagePipelines

CanManagePipelines checks for "orguser.org <> None" so you will be able to remove those checks from client_controller

PR #40, opened 2023-04-11
14 encrypt credentials to target warehouse before storing in database

Ishankoradia:
Keeping this open. We can merge this when we finalize on the encyprtion tool we will use.

Right now we are using kms encryption using a single key for all organizations


fatchat (commented):
this is nice, but a couple comments:
- when going between bytes and strings you should use encode() and decode() with a codec like utf8
- if an aws api call fails (or any api call for that matter), if you're not going to pass the error message down the call stack you should at least log it

Ishankoradia:
> this is nice, but a couple comments:
> 
> * when going between bytes and strings you should use encode() and decode() with a codec like utf8
> * if an aws api call fails (or any api call for that matter), if you're not going to pass the error message down the call stack you should at least log it

Got it. I have made the changes accordingly.

fatchat (approved):
won't merge at this time

fatchat:
this will be covered by PR #52

PR #41, opened 2023-04-11
enhancement - take server port from env and load env as soon as serve…

Ishankoradia:
Loading env as soon as the server starts. Will prevent us from using load_env() in every file.

Taking django server port from env helps since airbyte and other things also run on 8000


fatchat (changes requested):
hmm this won't work... runserver can be passed a port as a command line parameter

and when we run in production we will use gunicorn not manage.py, so if you want to load_dotenv you'll have to do it somewhere else

Ishankoradia:
I see, got it.

Issue #7, opened 2023-03-28
Break up clientapi into multiple controllers


fatchat:

fatchat:
As of March 28, clientapi has functionality for

- Client and User onboarding
- Airbyte setup for a client
- dbt setup for a client
- Prefect block creation for a client (Airbyte Connections and dbt Core)

... some of this is still in the branch 6-prefect-integration, so this issue can't be started until that branch is merged into main

To aid in maintainability, these different functionality groups could be split up into their own files, each referenced from urls.py

fatchat:
clientapi.py is now api/client.py

PR #44, opened 2023-04-14
clientapi into multiple controllers

Abhishek-N:
Fix: #7


ddpui/urls.py
fatchat, line 1: make the match patterns as long as possible i.e.
`path("api/airbyte/", airbyteapi.urls)`
etc

you can do it for airbyte, dbt and prefect without changing any endpoints

for the user_org_api we don't have a single prefix right now, so leave the pattern as "api/" and leave it at the end of the list. next week we can figure that out - possibly by splitting up user and org

fatchat (approved):

PR #45, opened 2023-04-17
standardizations + schemas with camel case + database field names with snake case + airbyte crud apis for server and connection block

Ishankoradia:


fatchat (changes requested):
camelCasing
$ref to find connection block having server id

todo: fix "blockid" in airbyteapi

fatchat (approved):

PR #51, opened 2023-04-20
24 api for prefect flow runs

Ishankoradia:


ddpui/models/org.py
fatchat, line 1: very minor request - but can you rename to OrgDataFlow? just want to introduce terminology to distinguish our flows from Prefect's flows... so we can call DDP flows "Data Flows"

ddpui/ddpprefect/prefect_service.py
```diff
@@ -308,15 +304,63 @@ def delete_dbt_core_block(block_id):
...
+        "sort": "START_TIME_DESC",
+        "deployments": {"id": {"any_": [deployment_id]}},
+        "flow_runs": {
+            "operator": "and_",
+            "state": {"type": {"any_": [FLOW_RUN_COMPLETED, FLOW_RUN_FAILED]}},
+        },
+    }
+
+    if limit:
+        query["limit"] = limit
```
fatchat, line 320: there's no "offset" field in the query?

fatchat (commented):
nice, i don't think it will take long to move the deployment / cron scheduling over to the prefect-proxy application

fatchat (approved):
merging so we can pull these changes into 19-dbt_cli_profile-for-bigquery

PR #53, opened 2023-04-24
update tests and remove older tests

Abhishek-N:


fatchat (changes requested):
tests passed on my machine when i ran `pytest` twice in succession

@Abhishek-N i see that there are no passwords in the source code (which is good) but you should also not put the other config info (host, user) in... a potential attacker now needs to guess only one thing instead of three. can you read them from ENV instead?

Abhishek-N:
yes missed that, its updated now @fatchat

PR #65, opened 2023-05-03
64 restrict airbyte destinations to bigquery and postgres

fatchat:
if `AIRBYTE_DESTINATION_TYPES` exists in `.env`, treat it as a comma-separated string of allowed destinations

e.g.
    `AIRBYTE_DESTINATION_TYPES="Postgres,BigQuery"`


ddpui/api/client/airbyte_api.py
```diff
@@ -268,6 +269,9 @@ def get_airbyte_destination_definitions(request):
         raise HttpError(400, "create an airbyte workspace first")
 
     res = airbyte_service.get_destination_definitions(orguser.org.airbyte_workspace_id)
```
siddhant3030, line 271: why don't you use this directly?  org = request.orguser.org and then use org everywhere

ddpui/api/client/airbyte_api.py
```diff
@@ -268,6 +269,9 @@ def get_airbyte_destination_definitions(request):
...
+    allowed_destinations = os.getenv('AIRBYTE_DESTINATION_TYPES')
+    if allowed_destinations:
+        res = [destdef for destdef in res if destdef['name'] in allowed_destinations.split(',')]
```
siddhant3030, line 274: I would write this something like this 

if allowed_destinations:
        destination_definitions = [d for d in destination_definitions if d['name'] in allowed_destinations.split(',')] 
        
and we can change the above res to destination_definitions

ddpui/api/client/airbyte_api.py
fatchat: i could but it hardly makes a difference... readability is the same, i doubt there is a performance difference and even if there is this is not meant to be high-performant code. it's the kind of change that if someone else did it i would roll my eyes but not stop them

ddpui/api/client/airbyte_api.py
fatchat: so you would rename `res` to `destination_definitions`? you could... all the other functions have `res` though

`res` is not as meaningful i agree, but if you change it in this function you should change it to more meaningful names everywhere else in the file

siddhant3030: Yes ideally we should do that but not on priority right now. I'll try to get involved in the PR review more often now.

Ishankoradia (approved):

PR #67, opened 2023-05-03
setup airbyte workspace when creating the org

fatchat:
instead of requiring the client to POST to /airbyte/workspaces/ , do the workspace setup during POST /organizations/

### DIFF
--- a/ddpui/ddpairbyte/airbytehelpers.py
+++ b/ddpui/ddpairbyte/airbytehelpers.py
@@ -1,0 +1,50 @@
+from django.utils.text import slugify
+from ddpui.ddpairbyte import airbyte_service
+from ddpui.ddpairbyte.schema import AirbyteWorkspace
+from ddpui.ddpprefect import prefect_service
+from ddpui.ddpprefect import AIRBYTESERVER
+from ddpui.models.org import OrgPrefectBlock
+from ddpui.utils.ddp_logger import logger
+
+def setup_airbyte_workspace(wsname, org) -> AirbyteWorkspace:
+    """creates an airbyte workspace and attaches it to the org
+    also creates an airbyte server block in prefect if there isn't one already
+    we don't need to update any existing server block because it does not hold
+    the workspace id... only the connection details of the airbyte server
+    """
+
+    workspace = airbyte_service.create_workspace(wsname)
+
+    org.airbyte_workspace_id = workspace["workspaceId"]
+    org.save()
+
+    # Airbyte server block details. prefect doesn't know the workspace id
+    block_name = f"{org.slug}-{slugify(AIRBYTESERVER)}"
+    display_name = wsname
+
+    airbyte_server_block_id = prefect_service.get_airbyte_server_block_id(
+        block_name
+    )
+    if airbyte_server_block_id is None:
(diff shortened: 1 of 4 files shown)


ddpui/ddpairbyte/airbytehelpers.py
```diff
@@ -0,0 +1,50 @@
...
+def setup_airbyte_workspace(wsname, org) -> AirbyteWorkspace:
+    """creates an airbyte workspace and attaches it to the org
+    also creates an airbyte server block in prefect if there isn't one already
+    we don't need to update any existing server block because it does not hold
+    the workspace id... only the connection details of the airbyte server
+    """
+
+    workspace = airbyte_service.create_workspace(wsname)
+
+    org.airbyte_workspace_id = workspace["workspaceId"]
```
siddhant3030, line 18: instead of calling workspace["workspaceId"] every time store it in the variable 

 workspace_id = workspace["workspaceId"]

ddpui/ddpairbyte/airbytehelpers.py
```diff
@@ -0,0 +1,50 @@
...
+    display_name = wsname
+
+    airbyte_server_block_id = prefect_service.get_airbyte_server_block_id(
+        block_name
+    )
+    if airbyte_server_block_id is None:
+        airbyte_server_block_id = prefect_service.create_airbyte_server_block(
+            block_name
+        )
+        logger.info(airbyte_server_block_id)
```
siddhant3030, line 32: add some comment to this 

something like this
logger.info(f"Created Airbyte server block with ID {airbyte_server_block_id}")

siddhant3030 (dismissed):

ddpui/ddpairbyte/airbytehelpers.py
fatchat: "every time"... you mean in the two places in this function??

ddpui/ddpairbyte/airbytehelpers.py
fatchat: done. our log messages are kinda here and there right now, we should set some standards and practices for logging. maybe you can do that... i'll open an issue for you

ddpui/ddpairbyte/airbytehelpers.py
siddhant3030: yes it's just better because think about this every time the code call it will call the function so instead of that we can store it in the variable and reference that value every time. Let me know if this is not clear.

fatchat: but it's not calling a function, it's doing a dictionary lookup

siddhant3030: workspace = airbyte_service.create_workspace(wsname) 

so this is returning the dictionary right so every time you're calling this workspace["workspaceId"] it's calling the function again? or am I wrong?

fatchat: the function is called once, the result of the function stored in a dictionary, the dict can be accessed as many times as you like the function will not be called again

You can verify this for yourself:

image

Ishankoradia (approved):

PR #103, opened 2023-05-19
Add workflow

Abhishek-N:
Fix: #102 

1. created a workflow
2. added host key to known hosts
3. ssh that key
4. port forwarded to 8000
5. created env ad saved credentials
6. run tests


fatchat (approved):
what an excellent idea!

PR #130, opened 2023-06-05
ignoring tests from codecov

siddhant3030:
The coverage is at 73% right now for one file which is test_airbyte_service. If there are no test cases for other files then it will not show all the coverage. So I'm keeping test coverage for 85% and we have to push our test cases to match that. 

let's ignore the build for sometime


fatchat:
`CONFLICT (directory rename split): Unclear where to rename ddpui/tests/unit_integration_tests to; it was renamed to multiple other directories, with no destination getting a majority of the files.
Automatic merge failed; fix conflicts and then commit the result.
`

when i tried to merge `main` into this branch on my machine

fatchat (approved):

PR #141, opened 2023-06-07
Test cases for airbyte service 

siddhant3030:
- We need to insure for each function we're passing the right values before it calls the API.


fatchat:
I don't think anyone is passing non-strings as workspace names... but we can make these checks useful by ensuring that the passed names meet airbyte's requirements, starting with the length

fatchat (approved):

PR #183, opened 2023-06-23
179 edit airbyte connection and reset connection apis

Ishankoradia:


ddpui/ddpairbyte/airbyte_service.py
```diff
@@ -632,79 +632,65 @@ def create_connection(
...
-    }
+    # update the name
+    if connection_info.name:
+        current_connection["name"] = connection_info.name
+
+    # update the destination schema
+    if connection_info.destinationSchema:
+        current_connection["namespaceDefinition"] = "customformat"
+        current_connection["namespaceFormat"] = connection_info.destinationSchema
+
```
fatchat, line 659: and if no `destinationSchema` in (the new) `connection_info`, should we not reset the `destinationSchema` which is potentially in `current_connection`?

fatchat (approved):
everything looks fine but please see my question in `update_connection`

PR #194, opened 2023-06-25
use the actual blockname when creating the manual-sync dataflow

fatchat:


Ishankoradia:
@fatchat I see we are cleaning up the name in the prefect proxy and not using that cleaned name to create deployment & set its parameters. Nice catch !!

Ishankoradia (approved):

PR #211, opened 2023-07-05
serve dbt docs

fatchat:


ddpui/html/docs.py
```diff
@@ -0,0 +1,30 @@
...
+        response = HttpResponse(html)
+        # the only valid values for x-frame-options are "deny" and "sameorigin", both
+        # of which will stop the iframe from rendering the docs
+        # removing the header causes it to be set to "deny" by django
+        # but if we set it to an invalid value, it makes its way to the browser where
+        # it is ignored
+        response.headers["X-Frame-Options"] = "ignore"
+        response.headers[
+            "Content-Security-Policy"
+        ] = "frame-src localhost:8002 ddpapi.projecttech4dev.org;"
```
Ishankoradia, line 29: should we take the api domain(s) from .env ? it will be easier when pushing it to a different machine

Ishankoradia (approved):

PR #214, opened 2023-07-05
update response

Abhishek-N:


fatchat (changes requested):
please check len(exc.args) and len(exc.args[0]) otherwise it will crash and we will look foolish!

PR #300, opened 2023-08-26
298 endpoint for prefect notifications

fatchat:


fatchat:
@Ishankoradia for your comments

PR #325, opened 2023-09-09
Job tracking

fatchat:


fatchat:
todo:
- end states in webhook
- expire the locks after one hour

fatchat:
when we deploy, we will run `python manage.py writedataflowblocks`

fatchat:
@Ishankoradia we can discuss before you merge

ddpui/management/commands/writedataflowblocks.py
```diff
@@ -0,0 +1,28 @@
...
+        parser.add_argument("--org-slug")
+
+    def handle(self, *args, **options):
+        """filters on --org-slug if provides"""
+        q_orgdataflow = OrgDataFlow.objects.all()
+        if options["org_slug"]:
+            q_orgdataflow = q_orgdataflow.filter(org__slug=options["org_slug"])
+
+        for odf in q_orgdataflow:
+            write_dataflowblocks(odf)
```
Ishankoradia, line 28: can we please not use 3 letter acronyms for vars instead of the full/partial name. breaks the readability. we have been doing very well with var names @fatchat

fatchat: But we love TLAs in the NGO field

fatchat:
we need to lock blocks for scheduled pipelines, since these are triggered by prefect we can't do this before they start so we listen for the state change to "Pending" in our webhook and lock over there

this state change will also be triggered for flows we have started manually, in those cases we acquire the locks before submitting the job to prefect so the webhook will fail to acquire them. since this is a-ok we ignore this failure

`/flows/dbt_run/` also needs block locking

Ishankoradia (approved):

PR #329, opened 2023-09-13
Last sync

fatchat:


fatchat:
@Ishankoradia can you make the webapp change where "last manual sync" is replaced by "last sync" thanks

Ishankoradia:
yes will do

Ishankoradia (approved):

PR #346, opened 2023-10-25
Warehouse apis

Abhishek-N:


fatchat (changes requested):
those four functions are 90% the same... ideal case for refactoring

on error please raise HttpError, don't silently log (we are not even looking at these logs...) and return an empty response

my vscode settings show me the double imports at the top, and tell me when my functions require docstrings... please make sure you have the same settings, i think i got mine from ishan you can ask him

requirements.txt
```diff
@@ -38,6 +38,7 @@ coverage==7.2.5
 croniter==1.3.8
 cryptography==39.0.2
 dateparser==1.1.7
+dbt-automation @ git+https://github.com/DalgoT4D/dbt-automation.git@33bf2d8778b2995bc6809496e29c1efb7ee0b345
 dbt-core==1.4.5
```
Ishankoradia, line 42: I think this should be `dbt-automation 0.1`, thats what it shows in my local django setup

Abhishek-N: throws me an error when i do this
```ERROR: Could not find a version that satisfies the requirement dbt-automation==0.1 (from versions: none)
ERROR: No matching distribution found for dbt-automation==0.1
```

fatchat (approved):

PR #356, opened 2023-11-08
fix - guest_token api request requires cookies to be set

Ishankoradia:


fatchat (approved):
getting the embed_token now

PR #373, opened 2023-11-20
368 refresh schema

fatchat:


Ishankoradia (approved):

Ishankoradia:
are we not working on this ? @fatchat  . Maybe we can take help from the C4GT folks.

fatchat:
Yes if we do they we'll make a new issue

PR #375, opened 2023-11-21
Parameterize dbt and remove blocks

Ishankoradia:


ddpui/api/client/airbyte_api.py
fatchat, line 1: in `post_airbyte_connection_reset_v1` why are we bothering to fetch the manual-sync deployment and verifying that its connection_id is the same? once we verify (on top) that the connection_id passed actually belongs to this client, can't we go ahead and reset right then?

fatchat: same question for update

fatchat (approved):

PR #405, opened 2023-12-31
User permissions

fatchat:
User permissions to 
- view / create / update / delete
- warehouses / sources / connections / users


fatchat:
pending:
- migrations
- tests

...and at least one use-case

PR #456, opened 2024-02-09
450 api to create a local dbt project

Abhishek-N:


fatchat:
two comments:
1. the `dbt` executable should be located according to the logic in `gather_dbt_project_params()`
2. till now we have not stored the dbt profiles in `~/.dbt/` ... instead we store them in prefect and prefect creates a `profiles.yml` in the dbt repo dir

fatchat (approved):

PR #471, opened 2024-02-16
Local dbt setup changes

Ishankoradia:


ddpui/ddpdbt/dbt_service.py
```diff
@@ -2,48 +2,107 @@
...
+    #     if shellblock.block_name.find("-git-pull") > -1:
+    #         try:
+    #             prefect_service.delete_shell_block(shellblock.block_id)
+    #         except Exception:  # pylint:disable=broad-exception-caught
+    #             pass
+    #         shellblock.delete()
+
+    # remove transform tasks
+    org_tasks_delete = []
+    for org_task in OrgTask.objects.filter(org=org, connection_id__isnull=True).all():
```
fatchat, line 51: shouldn't we filter on orgtask.task.type instead

ddpui/ddpdbt/dbt_service.py
```diff
@@ -97,7 +159,20 @@ def setup_local_dbt_workspace(org: Org, project_name: str, default_schema: str)
         logger.error(f"dbt init failed with {e.returncode}")
         return None, "Something went wrong while setting up workspace"
 
-    logger.info(f"starting to setup local dbt workspace at {project_dir}")
+    # setup packages.yml
+    dbtpackages_filename = Path(dbtrepo_dir) / "packages.yml"
+    with open(dbtpackages_filename, "w", encoding="utf-8") as dbtpackgesfile:
```
fatchat, line 164: even here we could copy a file from assets/

ddpui/ddpdbt/dbt_service.py
```diff
@@ -2,48 +2,107 @@
...
+            > 0
+        ):
+            raise Exception(f"{str(org_task)} is being used in a deployment")
+        org_tasks_delete.append(org_task)
+
+    logger.info("deleting orgtasks")
+    for org_task in org_tasks_delete:
+        for dataflow_orgtask in DataflowOrgTask.objects.filter(orgtask=org_task).all():
+            dataflow_orgtask.dataflow.delete()
+        org_task.delete()
```
fatchat, line 65: shouldn't we also delete the dataflow_orgtask

fatchat (changes requested):

ddpui/ddpdbt/dbt_service.py
Ishankoradia: that will happen automatically right ? datafloworgtask has `on_delete=models.CASCADE`

ddpui/ddpdbt/dbt_service.py
Ishankoradia: Okay sure. Let me make that change.

fatchat (approved):

PR #477, opened 2024-02-19
469 add custom dbt tasks to a pipeline

Ishankoradia:


ddpui/management/commands/set_uuids_in_orgtask_deployment.py
```diff
@@ -0,0 +1,206 @@
...
+                logger.info("skipping to next loop")
+                continue
+
+            # add uuid to each task config in params
+            params = deployment["parameters"]
+            if "config" in params and "tasks" in params["config"]:
+                for task_config in params["config"]["tasks"]:
+                    if "slug" in task_config:
+                        org_task = OrgTask.objects.filter(
+                            org=org, task__slug=task_config["slug"]
```
fatchat, line 77: over here... for the dbt orgtasks do you want to filter on `parameters__in=["", "{}"]` just in case some client has created a custom orgtask

ddpui/management/commands/set_uuids_in_orgtask_deployment.py
```diff
@@ -0,0 +1,206 @@
...
+                            task_config["orgtask_uuid"] = str(org_task.uuid)
+                            logger.info(
+                                f"updated task config with uuid for {task_config['slug']}"
+                            )
+                logger.info(
+                    f"updated deployment params for {dataflow.name}|{dataflow.deployment_id}"
+                )
+            else:
+                logger.info(
+                    f"no tasks key found in deployment params for {dataflow.deployment_id}"
```
fatchat, line 96: we upgraded all of them right? so over here you can raise an exception

ddpui/management/commands/set_uuids_in_orgtask_deployment.py
```diff
@@ -0,0 +1,206 @@
...
+
+            # assert if the uuid is present in each task config of each deployment
+            try:
+                deployment = prefect_service.get_deployment(dataflow.deployment_id)
+                params = deployment["parameters"]
+                if "config" in params and "tasks" in params["config"]:
+                    for task_config in params["config"]["tasks"]:
+                        if "orgtask_uuid" not in task_config:
+                            logger.info(
+                                f"uuid not found in task config {task_config['slug']} of deployment with id {dataflow.deployment_id}"
```
fatchat, line 128: this should also be an exception right

ddpui/management/commands/set_uuids_in_orgtask_deployment.py
```diff
@@ -0,0 +1,206 @@
...
+                    else:
+                        logger.error(
+                            f"Did not find orgtask_uuid in task config of deployment for slug {task_config['slug']} in {dataflow.deployment_id}"
+                        )
+                logger.info(
+                    f"updated deployment params seq for {dataflow.name}|{dataflow.deployment_id}"
+                )
+            else:
+                logger.info(
+                    f"no tasks key found in deployment params for {dataflow.deployment_id}"
```
fatchat, line 189: same, raise

ddpui/management/commands/set_uuids_in_orgtask_deployment.py
```diff
@@ -0,0 +1,206 @@
...
+                                orgtask=org_task, dataflow=dataflow
+                            ).update(seq=task_config["seq"])
+                        else:
+                            logger.error(
+                                f"Did not find orgtask with uuid {task_config['orgtask_uuid']} in task config of deployment for slug {task_config['slug']} in {dataflow.deployment_id}"
+                            )
+
+                    else:
+                        logger.error(
+                            f"Did not find orgtask_uuid in task config of deployment for slug {task_config['slug']} in {dataflow.deployment_id}"
```
fatchat, line 182: same, raise

ddpui/management/commands/set_uuids_in_orgtask_deployment.py
Ishankoradia: I mean we can but i dont want it to stop from going through other dataflows/deployments. What do you think ?

> this should also be an exception right

fatchat (approved):

PR #510, opened 2024-03-19
datatypes for postgres and bigquery

Abhishek-N:


ddpui/api/transform_api.py
```diff
@@ -573,3 +573,18 @@ def delete_operation(request, operation_uuid):
     dbtautomation_service.delete_dbt_model_in_project(dbt_operation.dbtmodel)
 
     return {"success": 1}
+
+
+@transformapi.get('/dbt_project/data_type/', auth=auth.CanManagePipelines())
+def get_postgres_datatypes(request):
```
fatchat, line 579: this is only for postgres?

ddpui/api/transform_api.py
```diff
@@ -576,15 +576,15 @@ def delete_operation(request, operation_uuid):
...
     org = orguser.org
 
     org_warehouse = OrgWarehouse.objects.filter(org=org).first()
     if not org_warehouse:
         raise HttpError(404, "please setup your warehouse first")
     
-    wclient = _get_wclient(org_warehouse)
-    data_types = wclient.get_column_data_types()
+    data_types = dbtautomation_service.warehouse_datatypes(org_warehouse)
     return Response(data_types)
```
fatchat, line 589: we usually return without wrapping in `Response`... i don't know what this does, whether it makes a difference to the response, or whether it is good style... all i know is that we don't explicitly wrap our responses in `Response` objects

fatchat (approved):

PR #531, opened 2024-04-05
Setup rbac stuff

Ishankoradia:


ddpui/management/commands/role_based_access.py
fatchat, line 1: umm what

ddpui/core/orguserfunctions.py
```diff
@@ -275,6 +277,86 @@ def invite_user(orguser: OrgUser, payload: InvitationSchema):
...
+        org=orguser.org, user__email__iexact=invited_email
+    ).exists():
+        return None, "user already has an account"
+
+    invited_role = Role.objects.filter(uuid=payload.invited_role_uuid).first()
+    if not invited_role:
+        return None, "Invalid role"
+
+    # user can only invite a role equal or lower to their role
+    if invited_role.level > orguser.new_role.level:
```
fatchat, line 298: the proper check is whether the inviter has the correct permission to invite another user

ddpui/auth.py
```diff
@@ -38,6 +46,32 @@ def authenticate(self, request, token):
...
+        @wraps(api_endpoint)
+        def wrapper(request, **kwargs):
+            # request will have set of permissions that are allowed
+            # check if permission_slug lies in this set
+            # throw error if nots
+            try:
+                if (
+                    request.permissions
+                    and len(
+                        set(permission_slugs).intersection(set(request.permissions))
```
fatchat, line 60: the request.permissions should be a subset of the permissions passed to the decorator

ddpui/core/orguserfunctions.py
fatchat: ok i see you're checking it in the middleware... in that case you can just remove this check

ddpui/management/commands/role_based_access.py
Ishankoradia: not clear what do you mean by that

ddpui/core/orguserfunctions.py
Ishankoradia: I think this should still be there.
 Even though the user has permissions to invite other users, an `account-manager` should not be able to invite a `super-user`.

ddpui/auth.py
Ishankoradia: Agreed. made the changes

ddpui/core/orguserfunctions.py
fatchat: oh like that... hmm all right

fatchat (approved):

PR #540, opened 2024-04-07
448 dump an orgs full config to file

Abhishek-N:


Abhishek-N:
https://docs.google.com/document/d/1w-6aP5ODfxF-bMj2T2VYtpdUStcVZeX7ZvoxBwROtHs/edit

ddpui/management/commands/dumpconfig.py
```diff
@@ -0,0 +1,142 @@
...
+            connection["destination"] = airbyte_service.get_destination(
+                org.airbyte_workspace_id, connection["destinationId"])
+            connection["source"] = airbyte_service.get_source(
+                org.airbyte_workspace_id, connection["sourceId"])
+            streams = connection["syncCatalog"]["streams"]
+            formatted_streams = []
+            for stream in streams:
+                formatted_stream = {
+                    "name": stream["stream"]["name"],
+                    "supportsIncremental": True,
```
fatchat, line 84: how come this is hardcoded to `True`?

fatchat (changes requested):
when dumping the pipelines, you should not dump the manual ones because those get created automatically when we create a connection. 

so don't dump the ones having `"dataflow_type": "manual"`

ddpui/management/commands/dumpconfig.py
Abhishek-N: oh we don't actually need to dump supportsIncremental, removed it

fatchat: ok ... can you explain what it is though, why it is there, why we don't need it

Abhishek-N: supportsIncremental is to check if the source supports incremental sync or not. it is not needed to create a connection because it takes it by default depending on the source. so we do not need to dump it

fatchat (approved):

PR #543, opened 2024-04-11
Custom macro unpivot

Ishankoradia:


fatchat:
so this copies all .sql files from the "assets" dir into each clients' dbt repos. but these assets are not under source control?

Ishankoradia:
> so this copies all .sql files from the "assets" dir into each clients' dbt repos. but these assets are not under source control?

They are but inside dbt-automation.
Check out the assets folder there.

fatchat (approved):

PR #562, opened 2024-04-22
559 distinguish between running and queued jobs   backend

fatchat:
WIP
To be merged/deployed with webapp https://github.com/DalgoT4D/webapp/pull/809


Ishankoradia:
@fatchat  

I see that there is now need to understand the status of a lock whether its running or queued.
Instead of maintaining flags like `isRunning` and `isCurrentJobQueued` on top sending the lock object `lock`. I have created an Enum of `TaskLockStatus`. A lock will now return 

```
{
    "lockedAt": "",
    "lockedBy": "",
    "flowRunId": "",
    "status": ""
}
```
The status is the status of flow run with  `flow_run_id`

Ishankoradia:
@fatchat ready for review whenever you get time

fatchat (commented):
Looking good

Ishankoradia (approved):

PR #569, opened 2024-04-25
fix: incorrect airbyte endpoints

Snehil-Shah:
Found incorrect endpoints

Expected endpoint: `/api/v1/operation/create/`
Actual endpoint: `api/v1//operation/create/`

### DIFF
--- a/ddpui/ddpairbyte/airbyte_service.py
+++ b/ddpui/ddpairbyte/airbyte_service.py
@@ -658,7 +658,7 @@ def create_normalization_operation(workspace_id: str) -> str:
 
     # create normalization operation
     res = abreq(
-        "/operations/create",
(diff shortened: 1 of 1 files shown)


fatchat:
in your webapps `.env`, make sure your `NEXT_PUBLIC_BACKEND_URL` doesn't have a trailing `/`

Snehil-Shah:
@fatchat yes, that's correctly setup. only these 2 endpoints have an extra `/` at the beginning. The abreq method already adds that `/` before the endpoint

fatchat (approved):

fatchat:
yes you are right

these endpoints are going away though, along with everything to do with "airbyte normalization"

thanks for finding it but we won't merge this PR

PR #573, opened 2024-04-28
572 test script for ui4t

fatchat:


fatchat:
merging this progress rather than keeping the branch open

PR #593, opened 2024-05-07
534 reset of connection gives no progress indicator in dalgo UI

Ishankoradia:


fatchat:
this is looking very nice @Ishankoradia . one request - could you make the create- and delete-connection functions resilient to prefect failures? like if they fail halfway through because prefect doesn't create / delete the deployment, the user should be able to press the button again and they should be able to run again

Ishankoradia:
> this is looking very nice @Ishankoradia . one request - could you make the create- and delete-connection functions resilient to prefect failures? like if they fail halfway through because prefect doesn't create / delete the deployment, the user should be able to press the button again and they should be able to run again

yes sure, let me know. good idea.

Ishankoradia:
@fatchat  done.

ddpui/ddpairbyte/airbytehelpers.py
```diff
@@ -243,6 +213,76 @@ def create_connection(org: Org, payload: AirbyteConnectionCreate):
...
+
+            sync_dataflow.reset_conn_dataflow = reset_dataflow
+            sync_dataflow.save()
+
+    except Exception as err:
+        # delete the airbyte connection; since the deployment didn't get created
+        logger.info("deleting airbyte connection")
+        airbyte_service.delete_connection(
+            org.airbyte_workspace_id, airbyte_conn["connectionId"]
+        )
```
fatchat, line 282: shouldn't we clean up any OrgTask / OrgDataFlowv1 which may have been created? or does the transaction.atomic automatically roll them back

ddpui/ddpairbyte/airbytehelpers.py
```diff
@@ -243,6 +213,76 @@ def create_connection(org: Org, payload: AirbyteConnectionCreate):
...
+            # create reset connection deplpyment & dataflow
+            org_task = OrgTask.objects.create(
+                org=org, task=reset_task, connection_id=airbyte_conn["connectionId"]
+            )
+
+            reset_dataflow: OrgDataFlowv1 = create_airbyte_deployment(
+                org, org_task, org_airbyte_server_block
+            )
+
+            sync_dataflow.reset_conn_dataflow = reset_dataflow
```
fatchat, line 274: shouldn't we also set the sync_dataflow within the reset_conn_dataflow

ddpui/ddpairbyte/airbytehelpers.py
```diff
@@ -340,12 +398,24 @@ def get_one_connection(org: Org, connection_id: str):
     if dataflow_orgtask is None:
```
fatchat, line 398: in the OrgTask query for this connection, don't we need to specify that we want the sync task

ddpui/ddpairbyte/airbytehelpers.py
fatchat: never mind, i see that you only look up the relationship in one direction

ddpui/ddpairbyte/airbytehelpers.py
Ishankoradia: Yes you are correct. Thanks for the catching that.

fatchat (approved):

PR #595, opened 2024-05-08
try a few times to fetch the cli profile

fatchat:


fatchat:
todo: look at postgres' "keep alive" settings and try adjusting to keep prefect from losing its connection

fatchat:
this seems to have stopped (?) since our prefect upgrade to 2.18.3

Ishankoradia:
I dont think so. Is there a way to keep track of this error (prefect postgres losing its connection again & again).

PR #597, opened 2024-05-09
script to move deployment to worker pool

Ishankoradia:


fatchat:
and this works? i was imagining it would be more complicated...

fatchat:
to make this change from agents to workers, you're going to

1. create a new work pool of type "process" named e.g. `default-worker-pool`
2. run this script with parameter `--workpool default-worker-pool`

shouldn't we also update the default pool name in `prefect-proxy:service.post_deployment_v1`
or will you be passing `payload.work_pool_name` from all callers

would you also record over here the pools with their names and types, for staging, beta and prod

thanks

fatchat (approved):

Ishankoradia:
All the steps in detail are here - https://github.com/DalgoT4D/platform_infra/issues/35

PR #605, opened 2024-05-15
Add support for Redis in docker

d-jeph:
# Support Redis on Docker

The goal for this PR is to add the support to run redis server as a docker container and allow the backend container to communicate with the redis server container.

The following changes have been made:

- Added a new file named `redis_client` in the utils folder. This file contains a class that is used to initialise a new redis server instance
- Updated all the scripts that interact with redis to use this new file to initialize a redis instance
- Updated the env.template file to add redis env variables: `REDIS_HOST` and `REDIS_PORT`
- Updated the docker compose files to add a `redis_server` service using the redis image from docker hub


ddpui/celery.py
```diff
@@ -5,7 +5,10 @@
 # Set the default Django settings module for the 'celery' program.
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "ddpui.settings")
 
-app = Celery("ddpui", backend="redis://localhost", broker="redis://localhost")
+REDIS_HOST = os.getenv("REDIS_HOST", "localhost")
+REDIS_PORT = int(os.getenv("REDIS_PORT", "6379"))
+
+app = Celery('ddpui', backend=f'redis://{REDIS_HOST}:{REDIS_PORT}', broker=f'redis://{REDIS_HOST}:{REDIS_PORT}')
```
yvonnegitau, line 11: Could you add a comment on what the backend does and what the broker does

ddpui/utils/redis_client.py
```diff
@@ -0,0 +1,18 @@
+import os
```
yvonnegitau, line 1: This is good

yvonnegitau (approved):
Looks good. Approving with some nit comment. I added the two docker files to enable one to run and test this

Docker/Dockerfile.dev.deploy
```diff
@@ -0,0 +1,17 @@
+FROM dalgo_main_image:0.1 as build
```
edodge, line 1: Can you add descriptions at the top of each Dockerfile so it's easy to understand the use case for each of them?

ddpui/utils/redis_client.py
```diff
@@ -0,0 +1,18 @@
...
+from redis import Redis
+
+
+class RedisClient:
```
edodge, line 5: Will this design create multiple connections from the app to Redis? Do we need to worry at all about managing multiple connections?

ddpui/utils/sendgrid.py
```diff
@@ -1,6 +1,8 @@
 import os
```
edodge, line 1: It looks like we are creating an email flow implementation that is specific to Sendgrid (using specific Sendgrid API calls and using Sendgrid-specific templates). I think a better design would be to only rely on Django's native email functionality. Then, at the deployment level the application can be configured to use whatever email service is desired. You can see [here](https://docs.sendgrid.com/for-developers/sending-email/django) how to configure the Django native email functionality to work with Sendgrid.

ddpui/celery.py
d-jeph: Added

ddpui/utils/redis_client.py
d-jeph: My thinking is it won't because the class method only creates a new connection instance if no previous instance exists else it just returns the existing instance.

ddpui/utils/sendgrid.py
d-jeph: I agree. This was Dalgo's design and implementation so we could flag it to them and then put it in our backlog as a tech debt task we can work on.

edodge (approved):

d-jeph:
merging redis changes

PR #608, opened 2024-05-17
Check if a github repo exists and return error 

mdshamoon:
## Summary

THis PR add the functionality to check whether a github repo exists or not before deleting the previous github repository in dbt transformation page.


mdshamoon:
> `check_repo_exists` doesn't work :-( tested with sneha's token on prod

Thanks for sharing this Rohit. Its seems the check that I added doesnt work with the .git extension. Have added a fix for it. Please test now.

fatchat:
yes now the api call works. BUT your replace logic doesn't:

```
'https://github.com/SNEHA-Org/dbt_sneha.git'.replace("github.com", "api.github.com/repos") => 'https://api.github.com/repos/SNEHA-Org/dbt_sneha.git'

'https://api.github.com/repos/SNEHA-Org/dbt_sneha.git'.replace(".git", "") => 'https://apihub.com/repos/SNEHA-Org/dbt_sneha'
```

if you switch the order of the replacements it works, but to avoid surprises can we use

```
import re
def convert_github_url(url):
    pattern = r'https://github.com/([^/]+)/([^/]+)\.git'
    replacement = r'https://api.github.com/repos/\1/\2'
    new_url = re.sub(pattern, replacement, url)
    return new_url
```

mdshamoon:
@fatchat this is ready for review

fatchat (approved):

PR #617, opened 2024-05-23
Feature/auto insights

Ishankoradia:
@fatchat  you can start looking at it. 
One thing i still need to figure out is Atomicity in redis. Redis doesn't do transactions way postgres does, so i have a sequence of commmands and if one fails - it will still run the others. And from what i have understand it doesn't have the concept of `rollback`.


ddpui/datainsights/generate_result.py
```diff
@@ -0,0 +1,424 @@
...
+            if redis.hsetnx(hash, key, json.dumps(payload)):
+                pass
+            else:
+                is_locked = True
+
+        if is_locked:
+            # poll for the lock till it releases basically till the query finishes running
+            current_lock = redis.hget(hash, key)
+            while current_lock is not None:
+                time.sleep(5)
```
fatchat, line 300: 5 is very long, set it to 1

Ishankoradia: Okay done

Ishankoradia:
@fatchat the test cases have passed but the coverage is low (<70%) and so the CI fails. Lets merge this, i will bring up the coverage in another PR.

PR #622, opened 2024-05-24
Backend on Docker

yvonnegitau:
# Backend on Docker

The goal of this task is to containerise the backend. This will simplify working with the backend and in the end simplify setting up Dalgo. Below are the changes made:

- Created two Dockerfiles, one for building the main image and another as an example on how to customize the main image for different deployments.
- Created a Docker Compose file. One for dev (`docker-compose.dev.yml`) environment where the developer uses postgres image as a database and the main (`docker-compose.yml`) using external database.
-  Added a mount folder. This is where all files that will be specific to a developer will be stored. An example is the `whitelist.py` file.
- Added documentation on how to run the backend using docker in Readme.
- We have added Redis change which will enable developer to add REDISHOST and REDISPORT

## Potential Breaking changes
Changes to the `gunicorn-log.conf` may affect applications already using it. Changed the path from absolute to relative considering developers may all not have the code in the same path.


d-jeph (approved):

d-jeph:
Looks good

PR #629, opened 2024-05-27
a centralized redis client for the django app; singleton class

Ishankoradia:


Ishankoradia:
@fatchat as we discussed. 
Backend docker PR can use this to setup the redis client and revert their changes

fatchat (approved):

PR #632, opened 2024-05-29
delete dbt orgtasks before deleting dbt workspace

fatchat:


ddpui/utils/deleteorg.py
```diff
@@ -150,6 +158,7 @@ def display_org(org: Org):
 
 def delete_one_org(org: Org):
     """delete one org"""
+    delete_dbt_orgtasks(org)
     delete_dbt_workspace(org)
```
Ishankoradia, line 162: @fatchat we are deleting the orgtasks in `delete_dbt_workspace` function.

fatchat: i think the order is wrong, workspace deletion fails at some point because some org tasks exist. the current script failed on staging without this change yesterday

Ishankoradia (approved):

PR #635, opened 2024-05-30
Get the credentials for airbyte from server block if feature flag is present

mdshamoon:
## Summary
This PR adds the following functionalities

1. Added a feature flag AIRBYTE_PROFILE. If its enabled for an organization, the credentials for Airbyte are fetched from Airbyte server block in prefect. Otherwise it defaults to env variables.
2. The library used for feature flags is Django-flags
3. We need to run the migration command after deploying the changes which creates a new table for feature flags.
4. I have enabled django admin portal (not sure if we would open it for production use as well). We can add/remove flags from the UI interface of django admin portal


ddpui/ddpairbyte/airbyte_service.py
```diff
@@ -29,13 +33,32 @@
...
     abver = os.getenv("AIRBYTE_SERVER_APIVER")
     token = os.getenv("AIRBYTE_API_TOKEN")
 
-    logger.info("Making request to Airbyte server: %s", endpoint)
+    if request is not None:
+        org_user = request.orguser
 
+        if flag_enabled("AIRBYTE_PROFILE", request_org_id=str(org_user.org.id)):
+            org_slug = org_user.org.slug
+            block_name = f"{org_slug}-{slugify(AIRBYTESERVER)}"
```
Ishankoradia, line 48: We should get the airbyte server block from the `OrgPrefectBlockv1` table and use the `block_name` stored there. Each org will have one.

ddpui/ddpairbyte/airbyte_service.py
```diff
@@ -29,13 +33,32 @@
...
     abport = os.getenv("AIRBYTE_SERVER_PORT")
     abver = os.getenv("AIRBYTE_SERVER_APIVER")
     token = os.getenv("AIRBYTE_API_TOKEN")
 
-    logger.info("Making request to Airbyte server: %s", endpoint)
+    if request is not None:
+        org_user = request.orguser
 
+        if flag_enabled("AIRBYTE_PROFILE", request_org_id=str(org_user.org.id)):
+            org_slug = org_user.org.slug
```
Ishankoradia, line 47: Can we add some `logger.info(` , to telll us whether its reading credentials from server block or from .env

ddpui/utils/flags.py
```diff
@@ -0,0 +1,7 @@
+from flags import conditions
+
+
+@conditions.register("org")
+def org_condition(org_id, request_org_id=None):
```
Ishankoradia, line 5: can we name the condition to `reigster("org_id")` ; so that someone updating things from admin panel knows it expects an id and not a org slug

do you think conditioning on `org_slug` would be more readable ? Just a thought, dont mind either.

ddpui/settings.py
```diff
@@ -67,8 +68,11 @@
     "ddpui",
     "django_prometheus",
     "django_extensions",
+    "flags",
 ]
 
+FLAGS = {"AIRBYTE_PROFILE": []}
+
```
Ishankoradia, line 75: can we add a comment here what is the purpose of this flag ?
Or if we can add description to a flag via the library which can be shown to the user on the admin panel.

ddpui/ddpairbyte/airbyte_service.py
mdshamoon: Updated

ddpui/ddpairbyte/airbyte_service.py
mdshamoon: Done.

ddpui/utils/flags.py
mdshamoon: I think slug would be more readable. Updated it.

ddpui/settings.py
mdshamoon: Done.

Ishankoradia (approved):

Ishankoradia:
Will merge once the ci finishes running

Ishankoradia (approved):

PR #636, opened 2024-05-31
create user with password

fatchat:


ddpui/management/commands/createorganduser.py
```diff
@@ -40,7 +41,15 @@ def handle(self, *args, **options):
 
         # fetch / create the User
         if not User.objects.filter(email=options["email"]).exists():
-            User.objects.create_user(email=options["email"], username=options["email"])
+            password = getpass.getpass("Enter the password for the OrgUser: ")
```
yvonnegitau, line 44: Is it possible to just pass it as an argument since we are automating the process? We can have it as a .env variable.

fatchat: nice idea, have added both

yvonnegitau (approved):
This looks good

PR #647, opened 2024-06-07
Name for orgdbtmodel

Abhishek-N:


ddpui/core/dbtautomation_service.py
```diff
@@ -168,7 +168,7 @@ def create_dbt_model_in_project(
             merge_input.append(
                 {
                     "input_type": source_model.type,
-                    "input_name": source_model.name,
+                    "input_name": source_model.name if source_model.type == 'model' else source_model.display_name,
                     "source_name": source_model.source_name,
```
Ishankoradia, line 172: @Abhishek-N can you explain why we need this change ?

Abhishek-N:
```
tables:
  - name: delhi
    identifier: delhi_india_measurement_form_2022
    description: One record per form data.
```
From code, here name is considered `display_name` and identifier is considered `name`. `display_name` is used for sources and `name` is used for models. Right now its working because both name and identifier are having same value

Ishankoradia (approved):

PR #649, opened 2024-06-07
edr as an orgtask

fatchat:
to run edr via prefect

### DIFF
--- a/ddpui/tests/core/test_orgtaskfunctions.py
+++ b/ddpui/tests/core/test_orgtaskfunctions.py
@@ -1,15 +1,37 @@
 import os
-import json
-from pathlib import Path
+import django
+from django.core.management import call_command
 from django.apps import apps
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "ddpui.settings")
+os.environ["DJANGO_ALLOW_ASYNC_UNSAFE"] = "true"
+django.setup()
+
 
 import pytest
 from ddpui.models.org import Org
 from ddpui.core.orgtaskfunctions import get_edr_send_report_task
 from ddpui.models.tasks import Task
-from ddpui.tests.api_tests.test_pipeline_api import seed_master_tasks_db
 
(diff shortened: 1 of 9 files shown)


ddpui/tests/core/test_orgtaskfunctions.py
```diff
@@ -0,0 +1,42 @@
...
+from ddpui.models.org import Org
+from ddpui.core.orgtaskfunctions import get_edr_send_report_task
+from ddpui.models.tasks import Task
+
+pytestmark = pytest.mark.django_db
+
+
+@pytest.fixture
+def seed_master_tasks():
+    """seeds the tasks table"""
```
Ishankoradia, line 16: we already have a common fixture to see master tasks `tests/api_tests/test_pipeline_api.py` -> `seed_master_tasks_db`

fatchat: yes i copied it over.. tried to import it and it wouldn't work :-(
that's why i modified it slightly to not double-add

Ishankoradia:
Some observations to keep in mind while this up
1. If its a new org, the endpoint `/elementary` will show `No Elementary report available`. 
2. Setting up edr for a new org consists of these steps
   1. Make sure the environment you are in has elementary installed.
   2. Create a directory inside the project folder called `elementary_profiles`. Generate a sample profile by running `dbt run-operation elementary.generate_elementary_cli_profile --profiles-dir profiles`. Use the output to create a new `profiles.yml` in `elementary_profiles` dir. Make sure you update the password. 
   3. Add to your `packages.yml` the dependency
    ```
     packages
          - package: elementary-data/elementary
            version: 0.15.2
    ```
   5. Add the below code in the `dbt_project.yml`. This is to define the schema where elementary can create its own models to track the test runs.
   ```
    models:
        elementary:
              +schema: "elementary"
    ```
   6. Run the initial elementary setup that create tables. Run `dbt deps &&  dbt run --select elementary`

PR #664, opened 2024-06-12
task for schema catalog

Abhishek-N:


ddpui/api/airbyte_api.py
```diff
@@ -632,9 +632,11 @@ def get_connection_catalog_v1(request, connection_id):
     if orguser.org.airbyte_workspace_id is None:
         raise HttpError(400, "create an airbyte workspace first")
 
-    res, error = airbytehelpers.get_connection_catalog(orguser.org, connection_id)
-    if error:
-        raise HttpError(400, error)
+    org_dict = orguser.org.__dict__
+    org_dict.pop('_state')
+    task = get_connection_catalog_task.delay(org_dict, connection_id)
+    res = task.get()
```
fatchat, line 638: `task.get` will block

fatchat: you need to return `task.id`. then you need to be able to poll the backend to find out if the task has completed. We typically do this using `TaskProgress`

fatchat (changes requested):

ddpui/api/airbyte_api.py
Abhishek-N: so we need to return both task.id and catalog response here

fatchat: no, the response won't be ready when `delay` returns, so you only have a task ID to return. look at how we set up a dbt workspace, or how we add custom airbyte connectors to a new airbyte workspace

PR #722, opened 2024-06-26
Used v2 of getting pipeline logs

mdshamoon:
## Summary

Used v2 of getting pipeline logs


fatchat:
is anyone using this function? if so then this change will break the caller ... didn't check but i assume the frontend calls this to get logs. if we merge this then we need to keep the front and back in sync which is slightly extra work

better to make a new function and call that one, then next week you can remove this old one

fatchat (approved):

PR #737, opened 2024-07-02
merged the logs summarization functions

Ishankoradia:
- Webhook not configured. Its an overkill at the moment, we will continue with polling.
- Have merged the two big function for log summarization.
- Also made some minor changes to the way we store store responses in our table
- Some migration changes


ddpui/api/pipeline_api.py
```diff
@@ -700,15 +703,21 @@ def get_prefect_flow_runs_log_history_v1(
 )
 @has_permission(["can_view_pipeline"])
 def get_flow_runs_logsummary_v1(
-    request, flow_run_id, regenerate: int
+    request, flow_run_id, task_id, regenerate: int = 0, request_uuid: str = None
```
fatchat, line 706: what's the request_uuid for?

fatchat (approved):

PR #749, opened 2024-07-09
log summarization updates

Ishankoradia:
- If there last session is running return the task_id
- save the task_ids in request_uuid.


fatchat:
instead of deleting the entry from the table in the event of an exception, can we save it with a FAILED flag

just in case we want to look at it later

Ishankoradia:
> instead of deleting the entry from the table in the event of an exception, can we save it with a FAILED flag
> 
> just in case we want to look at it later

Yes sure, let me make some changes

Ishankoradia:
Have made the changes @fatchat , makes sense to keep a session status

fatchat (approved):

PR #758, opened 2024-07-18
Json encode decimal

fatchat:


Ishankoradia:
@fatchat made some minor changes. Problem was only with the common query.

PR #774, opened 2024-07-29
Potential fix for task id bug

mdshamoon:
## Summary

Potential fix for task id bug

### DIFF
--- a/ddpui/api/airbyte_api.py
+++ b/ddpui/api/airbyte_api.py
@@ -667,7 +667,7 @@ def get_connection_catalog_v1(request, connection_id):
         return {"task_id": task_key, "message": "already running"}
 
     # ignore the returned celery task id
-    get_connection_catalog_task.delay(task_key, orguser.org.id, connection_id)
+    get_connection_catalog_task.delay(task_key, orguser.org, connection_id)
 
     return {"task_id": task_key}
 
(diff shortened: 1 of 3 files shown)


fatchat:
can ORM objects be passed to celery tasks? i remember this didn't used to be possible, which is why we pass database IDs to the tasks and make them retrieve the objects from the db themselves

even if this is possible, this PR won't do anything. what you want is to create the TaskProgress entry in the api handler, attach the celery task id and then return the task id to the caller

that way when the caller asks for an update on the progress of the task, the TaskProgress object with the task id actually exists

mdshamoon:
> can ORM objects be passed to celery tasks? i remember this didn't used to be possible, which is why we pass database IDs to the tasks and make them retrieve the objects from the db themselves
> 
> even if this is possible, this PR won't do anything. what you want is to create the TaskProgress entry in the api handler, attach the celery task id and then return the task id to the caller
> 
> that way when the caller asks for an update on the progress of the task, the TaskProgress object with the task id actually exists

Thanks for explaning @fatchat. I think I had a different understanding where I thought that the delay was due to the DB call made for getting the org with the help of the id passed.

mdshamoon:
Closing in favor of https://github.com/DalgoT4D/DDP_backend/pull/776

PR #795, opened 2024-08-06
Refresh elementary report through Prefect

mdshamoon:


ddpui/api/orgtask_api.py
```diff
@@ -277,6 +278,15 @@ def post_system_transformation_tasks(request):
     return {"success": 1}
 
 
+@orgtaskapi.get("elementary-lock/", auth=auth.CustomAuthMiddleware())
+@has_permission(["can_view_orgtasks"])
+def get_elemetary_task_lock(request):
+    """Check if the elementary report generation task is underway"""
+    org = request.orguser.org
+    org_task = get_edr_send_report_task(org)
```
fatchat, line 286: this raises an exception if the task doesn't exist, so make sure the client is prepared to catch it

fatchat (approved):

PR #796, opened 2024-08-06
Sentry bugs v2

Ishankoradia:


fatchat (approved):
our code is getting very complicated

Issue #804, opened 2024-08-08
Updated column is not propagated to the next operation


mdshamoon:
https://jam.dev/c/86c2f5fe-f7a1-41c9-a6de-3fd34f2346db

Issue #805, opened 2024-08-08
Edit operation bug


mdshamoon:
If an operation is in chain and you make edit to the first operation it breaks the input source model connection and nothing is updated in the destination model

https://jam.dev/c/19c4b6c6-c5cf-48a0-9cab-034480c81f22

PR #807, opened 2024-08-08
made sure the changes propagate and edit doens't lose the connection to source

Ishankoradia:
Fixes

https://github.com/DalgoT4D/DDP_backend/issues/804 

https://github.com/DalgoT4D/DDP_backend/issues/805

Tested manually for single input and multi input operations. (Might not have covered all operations).
Tested editing for first operation in the chain and an operation in middle of chain.

The way this will work is, if you edit one operation. The output_cols are propagated and available to the next one as source columns. However , the next operation might fail and the user is suppose to re-configure it based on the new column information in that op.

### DIFF
--- a/ddpui/api/transform_api.py
+++ b/ddpui/api/transform_api.py
@@ -301,12 +301,11 @@ def put_operation(request, operation_uuid: str, payload: EditDbtOperationPayload
 
     target_model = dbt_operation.dbtmodel
 
-    current_operations_chained = OrgDbtOperation.objects.filter(
-        dbtmodel=target_model
-    ).count()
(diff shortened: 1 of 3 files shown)


fatchat:
why the `depth` parameter, why not go till the end of the chain?

Ishankoradia:
> why the `depth` parameter, why not go till the end of the chain?

We can't because i dont know what config they want for the next set of operations. 
For example if we drop operation was edited with an additional col being dropped lets say `Indicator`. 
If the next operation was rename with config {"Indicator" -> "ind"} , now i dont know which columns. they want to rename after the editing. So i can't generate the output col for this op, to propagate.

fatchat (approved):

PR #811, opened 2024-08-11
Create missing DataflowOrgtasks, remove OrgDbt parameter from get_edr_send_report_task

fatchat:
Linking objects between the `edr` `OrgTasks` and the prefect deployments `OrgDataFlowv1` were missing. This script fixed that

Also noticed that `get_edr_send_report` doesn't need the `OrgDbt` parameter, since `org.dbt` exists. Removed it


Ishankoradia:
I see, makes sense. I should have done it in my PR. 
Thanks Rohit.

Ishankoradia (approved):

PR #818, opened 2024-08-16
Created task and added to redis so that we are able to fetch the task immediately on polling

mdshamoon:
Created task and added to redis so that we are able to fetch the task immediately on polling.
One of the fixes for https://github.com/DalgoT4D/webapp/issues/1052


fatchat (approved):
the redis key will be overridden in `SingleTaskProgress.__init__` but that should be okay

PR #824, opened 2024-08-23
Removed the deprecated function used in the airbytehelpers.setup_airb…

yashpal2104:
Removed "airbytehelpers.setup_airbyte_workspace_v1"

Since in this the API entrypoint is given so I removed assignment of workspace from "airbytehelpers.setup_airbyte_workspace_v1 "to "None" plz see if this works I could not understand how do I test this. Thanks

### DIFF
--- a/ddpui/api/airbyte_api.py
+++ b/ddpui/api/airbyte_api.py
@@ -446,7 +446,7 @@ def post_airbyte_workspace_v1(request, payload: AirbyteWorkspaceCreate):
     if orguser.org.airbyte_workspace_id is not None:
         raise HttpError(400, "org already has a workspace")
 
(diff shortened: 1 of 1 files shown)


fatchat:
??? how could this possibly work
the very next line will crash when you reference `workspace.workspaceId`

yashpal2104:
It did not raise any errors when I passed it through tests of deepsource ai

fatchat:
@yashpal2104 please read the code and understand what i'm saying. i'm not getting into online arguments with strangers

yashpal2104:
I am sorry @fatchat I don't intend to cause any inconveniences or arguments, I am extremely sorry if I made you think that way. I will surely try to solve this problem in the correct way. Thank you for pointing it out and giving your time to point out the error.

PR #826, opened 2024-08-23
Removed the deprecated function used in the airbytehelpers.setup_airb…

yashpal2104:
Removed "airbytehelpers.setup_airbyte_workspace_v1"

Since in this the API entrypoint is given so I removed assignment of workspace from "airbytehelpers.setup_airbyte_workspace_v1 "to "None" plz see if this works I could not understand how do I test this. Thanks


fatchat:
??

PR #829, opened 2024-08-23
Removed deprecated function in dashboard_api.py

yashpal2104:


fatchat:
too small for a PR

yashpal2104:
@fatchat is it too small of a PR if I have removed a deprecated functions(in the commit 1) and unused imports(in commit 2) because I have done that in my PR, if there is an error in removing the deprecated functions(in my commit 1) you can say so but I am unable to understand how this is a small contribution PR?

fatchat:
my mistake, i was looking only at the [first commit](https://github.com/DalgoT4D/DDP_backend/pull/829/commits/7d834372f31209effc4aa5ccd362a60ec87e026a)

now i see the entire PR i'm even more worried... what made you think this function was deprecated? we call it from our frontend

yashpal2104:
Isn't it mentioned here. get_prefect_dataflows_v1
Screenshot 2024-08-24 145356

PR #832, opened 2024-08-24
[442] - Pagination added for logs

huzaifmalik786:


fatchat:
depends on https://github.com/DalgoT4D/prefect-proxy/pull/154

Ishankoradia:
@fatchat @huzaifmalik786 is there a webapp PR coming for this.

huzaifmalik786:
@Ishankoradia , it has already been raised. Please see it here:
https://github.com/DalgoT4D/webapp/pull/1143

Thanks!

ddpui/api/pipeline_api.py
```diff
@@ -712,19 +712,14 @@ def get_prefect_flow_runs_log_history(
...
     flow_runs = prefect_service.get_flow_runs_by_deployment_id_v1(
         deployment_id=deployment_id, limit=limit, offset=offset
     )
 
-    if fetchlogs:
-        for flow_run in flow_runs:
-            logs_dict = prefect_service.get_flow_run_logs_v2(flow_run["id"])
-            flow_run["runs"] = logs_dict
-
     return flow_runs
```
Ishankoradia, line 723: Why did we remove this ? 
The `get_flow_run_logs_v2` is not just fetching the logs its also, fetching the graph of a `flow_run_id`. We need that to display the all the tasks inside the flow_run. That is why i asked you to create a new api on proxy to just fetch the and remove logs from it. 
I agree the name of the function is confusing a bit because first it used to only fetch logs but now its get the graph (or list of tasks inside a flow run also). 

Screenshot 2024-08-28 at 11 32 59

Ishankoradia:
@huzaifmalik786 works nicely.

For the ai summary, can you make sure we are fetching all logs in the api `v1/flow_runs/{flow_run_id}/logsummary`. 
It internally calls a function `summarize_logs` that in turn calls `get_flow_run_logs_v2` . maybe replace this with a common function that fetches all pages with some configurable limit. 

Thanks. Nice work.

ddpui/celeryworkers/tasks.py
```diff
@@ -762,6 +765,15 @@ def summarize_logs(
...
                 return
             task = dbt_tasks[0]
+            task['logs'] = []
+            limit = 0
+            while(True):
+                new_logs_set = get_flow_run_logs(flow_run_id, task_id, limit, offset=FLOW_RUN_LOGS_OFFSET_LIMIT)
+                task['logs']+=new_logs_set['logs']
+                if len(new_logs_set['logs']) == FLOW_RUN_LOGS_OFFSET_LIMIT:
+                    limit+=1
+                else:
```
Ishankoradia, line 775: @huzaifmalik786 looks like offset and limit have been switched ?

limit will increment by limit = limit + FLOW_RUN_LOGS_OFFSET_LIMIT

Ishankoradia: and we are not incrementing offset . logic is a bit off.

Ishankoradia (approved):

PR #840, opened 2024-08-27
819 add created at and updated at to all django tables

Ishankoradia:
Skipped
- `OrgTnC`
- `Role`, `Permission`, `RolePermission`
- `SyncStats`
- `Task`

The `auto_now` timestamp doesn't work when we update a model using `.update(some=some)`. Since in this, django does directly sql query to db. Timestamps will work with `.save()`


fatchat (approved):

Ishankoradia:
@fatchat merging this. Luckily we haven't use `.update()` a lot besides some management commands.

PR #841, opened 2024-08-28
Incorrect order of dbt tasks

Ishankoradia:
This PR does the following
- Addes `dbt seed` as a system generated task
- Removes `dbt docs generate` from current and future dataflows/pipelines
- Removes `dbt docs generate` from being used as the default task while adding a pipeline using `simple` strategy
- A script to find out if the deployment seq is messed up or not. If it is, pass `--fix` to fix it. The seq is checked based on the seq defined `ddpui/utils/constants.py TRANSFORM_TASKS_SEQ`. This is only for transform tasks. 
- Allow `dbt docs generate` to be added as a custom task.
- Corresponding webapp PR - https://github.com/DalgoT4D/webapp/pull/1158 

The original issue is not reproducible.

### DIFF
--- a/ddpui/api/orgtask_api.py
+++ b/ddpui/api/orgtask_api.py
@@ -50,11 +50,11 @@ from ddpui.utils import secretsmanager
 from ddpui.utils import timezone
 from ddpui.utils.helpers import map_airbyte_keys_to_postgres_keys
 from ddpui.utils.constants import (
-    TASK_DBTRUN,
     TASK_GITPULL,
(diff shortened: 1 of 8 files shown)


fatchat (approved):
this is fine... my only comment is that in `fix_transform_tasks_seq_dataflow` assumes that transform tasks always come last

which of course they do at this time

(btw nobody is calling that function)

Ishankoradia:
@fatchat the script is calling that function. 
I put it there so we can use shell fix stuff on the go if we come across any discrepancies.

PR #842, opened 2024-08-29
remove flags

Ishankoradia:


fatchat (approved):
then what did shamoon change

Ishankoradia:
> then what did shamoon change

not sure

PR #845, opened 2024-08-30
834 add poc demo attribute in org

Ishankoradia:


fatchat (approved):
Can you make a list of orgs whose types should be changed to `POC` once this is released

Ishankoradia:
Sure.

PR #849, opened 2024-09-02
Deleting warehouse bug

### LINKED ISSUE #836: deleting the warehouse and re-creating doesn't refresh the transformation prefect cli block

Ishankoradia:
Fix #836


ddpui/ddpairbyte/airbytehelpers.py
```diff
@@ -762,27 +770,210 @@ def update_destination(
...
+        try:
+            cli_block_response = prefect_service.update_dbt_cli_profile_block(
+                block_name=cli_profile_block.block_name,
+                wtype=warehouse.wtype,
+                credentials=dbt_creds,
+                bqlocation=bqlocation,
+                profilename=profile_name,
+                target=target,
+            )
+            cli_profile_block.block_id = cli_block_response["block_id"]
```
fatchat, line 911: we don't need this for updates right, block id and name won't change?

fatchat (approved):

PR #851, opened 2024-09-05
817 run up to down from a selected model

Ishankoradia:


ddpui/api/dbt_api.py
```diff
@@ -265,7 +266,7 @@ def post_run_dbt_commands(request):
     taskprogress.add({"message": "Added dbt commands in queue", "status": "queued"})
 
     # executes clean, deps, run
-    run_dbt_commands.delay(orguser.id, task_id)
+    run_dbt_commands.delay(orguser.id, task_id, payload.dict())
```
fatchat, line 269: does payload.dict() work if payload is None?

Issue #860, opened 2024-09-18
schedule large connection jobs over the weekend


Ishankoradia:
For connections syncing large data, certain jobs that might break the machine can be pushed off to the weekend.
This will only be for select connection_ids which can be called `large`
- [ ] Schema change
- When user opens the schema changes modal, he will see all the changes along with a message `Accepting these changes, will schedule a schema job on {Date}`
- As soon as user accepts the schema change updates, we will schedule the schema change job (update catalog + reset + sync ) on the nearest Sunday. 
- The schema changes will still appear in the `Pending Action` tab along with the datetime of when it is scheduled. 
- If before the next scheduled time, there is a new schema change and the user wants to accept those changes. He can do it and we will update the flow_run params & make sure it again scheduled on the nearest Sunday. 
- Store the `scheduled_at` and `flow_run_id` in the backend.

- [ ] Connection Reset
- When the user presses on the `Reset` connection button on the ingest page. A modal appears that asks for confirmation.
- The confirmation modal will also have the message `On clicking confirm, this will schedule a reset connection job on {Date}`
- Store the `scheduled_at` and the `flow_run_id` for the upcoming reset  job somewhere in the backend

@fatchat

nairabhishek73:
Schema Change: https://jam.dev/c/3dbdf287-ea6d-4316-8128-9bac2c8d9bb0
1) After I approve I still see it under pending actions, I can open it and then approve it again. 
- Once I approve it should indicate to me that its approved and scheduled
- If there is another schema change top the same connection it should show up separately as a pending action, if approved it should let me know that the other approval is overridden by the latest one and ask for confirmation
2) Relatedly, but perhaps a separate issue, when I click on view I should see the connection name on the pop up box and then the details of the schema changes within that connection.

Tried connection reset, I don;t think that's done yet.

PR #861, opened 2024-09-18
Schedule large connections jobs

Ishankoradia:
Feature https://github.com/DalgoT4D/DDP_backend/issues/860


Ishankoradia:
Corresponding frontend #861

ddpui/management/commands/mark-connections-as-large-enough.py
```diff
@@ -0,0 +1,49 @@
...
+    help = """This script is to figure out if the connection is large 
+    enough to schedule related schema change & reset jobs for later"""
+
+    def add_arguments(self, parser: CommandParser) -> None:
+        parser.add_argument("--org", type=str, help="org slug, optional", required=True)
+        parser.add_argument(
+            "--connection_id",
+            type=str,
+            help="connection_id, optional; if not passed all connections of the org will be considered",
+            required=False,
```
fatchat, line 25: let's make this a required parameter? the only time we would run the script without specifying a connection ID would be to create default `ConnectionMeta` objects, but the logic which accesses `schedule_large_jobs` works / should work without a `ConnectionMeta`

ddpui/utils/helpers.py
```diff
@@ -182,3 +184,25 @@ def convert_sqlalchemy_rows_to_csv_string(rows: list[dict]):
 def convert_sqlalchemy_rows_to_json_string(rows: list[dict]):
     """converts a list of sqlalchemy rows to a csv string"""
     return json.dumps(convert_to_standard_types(rows), indent=4)
+
+
+def get_schedule_time_for_large_jobs(
+    time: time = time(10, 0), weekday: int = calendar.SUNDAY
+) -> datetime:
+    """By default scheduled time will be Sunday 10am UTC"""
+    now = datetime.now(pytz.utc)
```
fatchat, line 193: can `now` be a parameter, will make it easier to write a test function

ddpui/ddpairbyte/airbytehelpers.py
```diff
@@ -500,14 +513,96 @@ def get_one_connection(org: Org, connection_id: str):
...
+    is_connection_large_enough = (
+        True == connection_meta.schedule_large_jobs if connection_meta else False
+    )
+
+    schedule_at = None
+    job: ConnectionJob = None
+    if is_connection_large_enough:
+        schedule_at = get_schedule_time_for_large_jobs()
+        # if there is a flow run scheduled , delete it
+        job = ConnectionJob.objects.filter(
```
fatchat, line 572: 1. the lines `job = ... if job: ...` can be pulled out of `if is_connection_large_enough`

thinking of a case where we update `is-large-enough` in the meta when there is already a job scheduled

3. after `delete_flow_run` can we update `job.flow_run_id = None` and save

thinking of a case when scheduling the next one fails after deletion succeeds

ddpui/ddpairbyte/airbytehelpers.py
```diff
@@ -1042,12 +1137,62 @@ def get_schema_changes(org: Org):
...
+                            "flow_run_id": change.schedule_job.flow_run_id,
+                            "job_type": change.schedule_job.job_type,
+                            "status": None,
+                            "state_name": None,
+                        }
+                        if change.schedule_job
+                        else None
+                    ),
+                    "is_connection_large": change.connection_id in large_connections,
+                    "next_job_at": get_schedule_time_for_large_jobs(),
```
fatchat, line 1172: can we put `if change.schedule_job else None` here as well? just so that in future we don't get confused when looking directly at the database

ddpui/ddpairbyte/airbytehelpers.py
```diff
@@ -1079,14 +1224,39 @@ def schedule_update_connection_schema(
...
+        )
+    else:
+        schedule_at = get_schedule_time_for_large_jobs()
+        # if there is a flow run scheduled , delete it
+        job = ConnectionJob.objects.filter(
+            connection_id=connection_id, job_type=UPDATE_SCHEMA
+        ).first()
+        if job:
+            try:
+                prefect_service.delete_flow_run(job.flow_run_id)
```
fatchat, line 1250: set `job.flow_run_id = None` and save, just in case there's a failure before the update below

ddpui/ddpairbyte/airbytehelpers.py
Ishankoradia: When do you do the schema update for the first for a large connection.
`change.schedule_job` will be `None` but frontend still needs to show in advance at what time this will be scheduled if they accept schema changes

ddpui/ddpairbyte/airbytehelpers.py
Ishankoradia: 1. we dont want to delete a job if the connections is not large enough. So how would taking it out help ? Correct if i misunderstood your comment

3. done

ddpui/ddpairbyte/airbytehelpers.py
fatchat: ah all right then

ddpui/ddpairbyte/airbytehelpers.py
```diff
@@ -499,14 +513,100 @@ def get_one_connection(org: Org, connection_id: str):
...
+    # if the connection is large enough, we will schedule a new flow run
+    # if the connection is not larged enough, we run it now
+    # either way we need to delete this job
+    job = ConnectionJob.objects.filter(
+        connection_id=connection_id, job_type=TASK_AIRBYTERESET
+    ).first()
+    if job:
+        try:
+            prefect_service.delete_flow_run(job.flow_run_id)
+            job.delete()
```
fatchat, line 581: we should set `job = None` after this no
otherwise below code will break

ddpui/ddpairbyte/airbytehelpers.py
```diff
@@ -1082,14 +1232,42 @@ def schedule_update_connection_schema(
...
+
+    # if there is a flow run scheduled , delete it
+    # if the connection is large enough, we will schedule a new flow run
+    # if the connection is not larged enough, we run it now
+    # either way we need to delete this job
+    job = ConnectionJob.objects.filter(connection_id=connection_id, job_type=UPDATE_SCHEMA).first()
+    if job:
+        try:
+            prefect_service.delete_flow_run(job.flow_run_id)
+            job.delete()
```
fatchat, line 1261: `job = None`

ddpui/utils/helpers.py
```diff
@@ -182,3 +184,28 @@ def convert_sqlalchemy_rows_to_csv_string(rows: list[dict]):
...
+        extra_delta = timedelta(hours=1)
+        time_of_day = None  # since it will time at now + extra delta
+
+    # Calculate the next weekday datetime
+    next_slot = now + timedelta(days=days_until_sunday) + extra_delta
+
+    # set the time
+    if time_of_day:
+        next_slot = datetime.combine(next_slot, time_of_day, tzinfo=pytz.utc)
+
```
fatchat, line 210: also... we should make sure `next_slot` is not in the past, and if it is we should push it by another week

Ishankoradia:
- Make sure the nightly job for schema change deletes the schema change as well as the prefect flow run

ddpui/utils/helpers.py
```diff
@@ -182,3 +184,28 @@ def convert_sqlalchemy_rows_to_csv_string(rows: list[dict]):
...
+    if days_until_sunday == 0:
+        extra_delta = timedelta(hours=1)
+        time_of_day = None  # since it will time at now + extra delta
+
+    # Calculate the next weekday datetime
+    next_slot = now + timedelta(days=days_until_sunday) + extra_delta
+
+    # set the time
+    if time_of_day:
+        next_slot = datetime.combine(next_slot, time_of_day, tzinfo=pytz.utc)
```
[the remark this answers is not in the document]
Ishankoradia: Thats the idea. We only use extra delta when its a sunday for which `time_of_day` passed might be inaccurate.

ddpui/utils/helpers.py
Ishankoradia: its not possible that next_slot is in past.

fatchat (approved):

PR #876, opened 2024-10-10
Feature/data insights out of ui4t

Ishankoradia:
Corresponding Webapp - https://github.com/DalgoT4D/webapp/pull/1181


ddpui/core/warehousefunctions.py
```diff
@@ -0,0 +1,76 @@
...
+
+    return convert_to_standard_types(data)
+
+
+def fetch_warehouse_tables(request, org_warehouse, cache_key=None):
+    """
+    Fetch all the tables from the warehouse
+    Cache the results
+    """
+    res = []
```
fatchat, line 57: just in case this takes a while, could we implement a lock here? i'm imagining users clicking the "sync sources" button repeatedly

maybe set some "in progress" flag and if the flag is set then return None

fatchat: or you can do it in the endpoint handler

Ishankoradia: This will only take a while the first time.

Ishankoradia:
All tests have passed, coverage is 1% low.

fatchat (approved):

PR #878, opened 2024-10-21
Backend docker changes

Ishankoradia:


Docker/Dockerfile.main
```diff
@@ -2,37 +2,40 @@
...
+RUN mkdir -p ${DATA_DIR}
+
 COPY requirements.txt /requirements.txt
 
 # install uv. This makes installation of packages faster compared to pip
 ADD https://astral.sh/uv/install.sh /install.sh
 
 #Install system dependencies. git is needed because there are some packages that are not available in PyPI
-RUN apt-get update && apt-get -y install libpq-dev gcc wget curl git \
+RUN apt-get update && apt-get -y install libpq-dev gcc wget curl git curl \
```
fatchat, line 23: `curl` appears twice

Docker/docker-compose.yml
```diff
@@ -9,77 +9,47 @@ services:
...
-      - DEMO_AIRBYTE_SOURCE_TYPES=${DEMO_AIRBYTE_SOURCE_TYPES}
-      - DEMO_SENDGRID_SIGNUP_TEMPLATE=${DEMO_SENDGRID_SIGNUP_TEMPLATE}
-      - DEMO_SUPERSET_USERNAME=${DEMO_SUPERSET_USERNAME}
-      - DEMO_SUPERSET_PASSWORD=${DEMO_SUPERSET_PASSWORD}
+    env_file:
+      - .env.docker
+    volumes:
+      - ${CLIENTS_DBT_MOUNT}:/data/clients_dbt
+      - ${DEV_SECRETS_MOUNT}:/data/secrets
+      - ${LOGS_MOUNT}:/usr/src/backend/ddpui/logs
```
fatchat, line 22: `/usr/src/backend/ddpui` ... what is this

Docker/docker-compose.yml
```diff
@@ -9,77 +9,47 @@ services:
...
+    command: worker
+    depends_on:
+      - backend
+      - redis_server
+    env_file:
+      - .env.docker
+    volumes:
+      - ${CLIENTS_DBT_MOUNT}:/data/clients_dbt
+      - ${DEV_SECRETS_MOUNT}:/data/secrets
+      - ${LOGS_MOUNT}:/usr/src/backend/ddpui/logs
```
fatchat, line 49: same q

Docker/Dockerfile.main
```diff
@@ -2,37 +2,40 @@
...
     && rm -rf /var/lib/apt/lists/* \
-    && apt-get clean 
-
-
+    && apt-get clean
 
-RUN mkdir -p ${HOME_DIR}/secrets \
-    && mkdir -p ${HOME_DIR}/dbt/dbt \
-    && mkdir -p ${HOME_DIR}/ddpui/logs 
+RUN mkdir -p ${DATA_DIR}/secrets \
```
fatchat, line 31: you need to create the `dbt` folder because you're going to `pip install` things in the next step, but you don't need to create `secrets` or `clients_dbt`. Docker will create them during the volume-mapping stage of the docker-compose

chat gpt said:
```
In Docker Compose, when you define a volume mapping like - host_path:/path/in/container, the container path (/path/in/container) does not need to exist in the image. Docker will create the entire directory structure in the container if it doesn't already exist, based on the structure specified in the volume mapping.
```

Docker/Dockerfile.main
Ishankoradia: you are right , thanks for pointing it out.

Docker/docker-compose.yml
Ishankoradia: its the `HOME_DIR`, defined at the top. 
I am not fussy about it, it can also be `/` or `/app` or anything. Its where we copy our code to.

Docker/Dockerfile.main
Ishankoradia: Agreed. I tested it out, it does create the directories.

fatchat (approved):

PR #880, opened 2024-10-30
making orgdbt's project_dir and dbt_venv relative

Ishankoradia:
@fatchat as we discussed, making paths in our db relative so that we have more flexibility and it will be easier to adopt the docker deployment


Ishankoradia:
All test cases have passed, codecov is under limit

fatchat (approved):

PR #884, opened 2024-11-08
update the prefect secret block when the github PAT is updated

fatchat:
Depends on https://github.com/DalgoT4D/prefect-proxy/pull/162

### DIFF
--- a/ddpui/api/dbt_api.py
+++ b/ddpui/api/dbt_api.py
@@ -15,7 +15,7 @@ from ddpui.celeryworkers.tasks import (
 from ddpui.models.tasks import TaskProgressHashPrefix
 from ddpui.utils.taskprogress import TaskProgress
 from ddpui.ddpdbt import dbt_service
-from ddpui.ddpprefect import DBTCLIPROFILE, prefect_service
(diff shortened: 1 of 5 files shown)


ddpui/api/dbt_api.py
```diff
@@ -70,6 +70,19 @@ def put_dbt_github(request, payload: OrgDbtGitHub):
     org.dbt.gitrepo_access_token_secret = payload.gitrepoAccessToken
     org.dbt.save()
 
+    if payload.gitrepoAccessToken is not None and payload.gitrepoAccessToken != "":
+        gitrepo_url = payload.gitrepoUrl.replace(
```
Ishankoradia, line 74: we need to create the secret block if the token is passed and the block does not exist

fatchat: good catch, i have changed the edit into an upsert please see

Ishankoradia: we also create the secret block in our db in `OrgPrefectBlockv1`. We should do a upsert in that django table also

fatchat: ok done

Ishankoradia (approved):

PR #887, opened 2024-11-12
Settings panel

himanshudube97:
I have a question regarding migration files. Should every small change in migrations be pushed to the main branch, or is there a better approach? I checked that Django provides a way to squash migrations. Do we use that?


fatchat:
we don't squash migrations, no. i'm not opposed to it, just we haven't so far

ddpui/utils/constants.py
Ishankoradia, line 1: we should not hard code the server & host for prefect and airbyte.

can we use from the `.env`

ddpui/api/org_preferences_api.py
```diff
@@ -0,0 +1,306 @@
...
+        return {"success": False, "error": "Organization not found"}
+
+    versions = []
+
+    # Airbyte Version
+    try:
+        airbyte_url = AIRBYTE_URL_TO_GET_VERSION
+        airbyte_response = requests.get(airbyte_url)
+        if airbyte_response.status_code == 200:
+            airbyte_data = airbyte_response.json()
```
Ishankoradia, line 258: we use airbyte_service to talk/call any apis of airbyte. lets move this as a helper there

ddpui/api/org_preferences_api.py
```diff
@@ -0,0 +1,306 @@
...
+@orgpreference_router.get("/toolinfo", auth=auth.CustomAuthMiddleware())
+def get_tools_versions(request):
+    orguser: OrgUser = request.orguser
+    org = orguser.org
+
+    try:
+        # Validate org_id
+        org_superset = OrgSupersets.objects.get(org=org)
+    except OrgSupersets.DoesNotExist:
+        return {"success": False, "error": "Organization not found"}
```
Ishankoradia, line 249: we should be raising http error with that message "Organization not found". You can look at some other api code files.

ddpui/api/org_preferences_api.py
```diff
@@ -0,0 +1,306 @@
...
+    """Gets preferences for an organization by org_id"""
+    orguser: OrgUser = request.orguser
+    org = orguser.org
+
+    try:
+        org_superset = OrgSupersets.objects.select_related(
+            "org",
+        ).get(org=org)
+
+    except OrgSupersets.DoesNotExist:
```
Ishankoradia, line 204: why are we doing select_related here ? we already have the org object at line 197

i think we can go away with the try except since we have a global try except in routers file that will catch all generic errors.

fatchat (approved):

PR #893, opened 2024-11-19
Replace reset with clear

Ishankoradia:
- Creating a new connection will now create a sync dataflow and a clear dataflow
- Added a script to add clear dataflow for current connections of current clients
- Update the sync history api to fetch jobs of both types `sync` & `reset_connection`

Once this is released, i will submit another PR to clear up objects in django and deployments related to reset deployment


ddpui/ddpairbyte/airbytehelpers.py
```diff
@@ -775,7 +795,9 @@ def get_sync_job_history_for_connection(
         return None, "connection not found"
 
     res = {"history": [], "totalSyncs": 0}
-    result = airbyte_service.get_jobs_for_connection(connection_id, limit, offset)
+    result = airbyte_service.get_jobs_for_connection(
+        connection_id, limit, offset, job_types=["sync", "reset_connection"]
```
fatchat, line 799: why don't we ask for `clear` over here?

fatchat:
> Once this is released, i will submit another PR to clear up objects in django and deployments related to reset deployment

maybe we should keep the reset task in the seed file

ddpui/ddpairbyte/airbytehelpers.py
Ishankoradia: Ahh yes

The clear jobs it seems come under `reset_connection`. Although the api documentation says `clear` is a configType but it throws an unknown error when i use it. 

However the above configuration, gives the correct results. But i have no way making out whether its a clear or reset. I dont know how airbyte shows it in its job history. (even airbyte queries on their UI with those configTypes)

fatchat (approved):

PR #895, opened 2024-11-19
Create Org plan, when an org is created.

himanshudube97:
1. Create or plan , when org is created.


ddpui/api/user_org_api.py
```diff
@@ -572,6 +574,11 @@ def post_organization_v1(request, payload: OrgSchema):
     # create a new orguser if the org is already there
     orguserfunctions.ensure_orguser_for_org(orguser, org)
 
+    # create a new orgplan
+    org_plan, error = orgfunctions.create_orgPlan(payload, org)
```
fatchat, line 578: don't mix snake_case with camelCase.. call it `create_org_plan`

ddpui/celeryworkers/tasks.py
```diff
@@ -963,15 +963,16 @@ def summarize_warehouse_results(
...
     roles_to_notify = [ACCOUNT_MANAGER_ROLE]
     days_before_expiry = 7
 
     for org in Org.objects.all():
         org_plan = OrgPlans.objects.filter(org=org).first()
         if not org_plan:
             continue
-
+        if org_plan and not org_plan.end_date:
+            continue
```
fatchat, line 975: add one more condition so that we only send emails to free trial accounts

himanshudube97: Added the condition.

ddpui/api/user_org_api.py
himanshudube97: Done

fatchat (approved):

PR #897, opened 2024-11-21
django admin app

Ishankoradia:
To login ; you will need to be a `is_superser` and `is_staff`

@fatchat see how this looks. We can also expose this on production. 

Have added llm session feedback


Ishankoradia:
run this to generate the static files under `static`

`python manage.py collectstatic  `

fatchat:
before we choose to use django views can we see how easy it will be to add charts

streamlit makes it very easy to do charts, tables, and form controls

Ishankoradia:
@fatchat lets do one chart (anyone that you like).
You can create on streamlit, i can do it on django. we cna compare notes.

PR #905, opened 2024-11-30
881 trigger various notifications using the current system in place

fatchat:
Notifications from the Prefect webhook handler

Long-running flow-runs are being handled by a Celery task


.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "editor.defaultFormatter": "ms-python.black-formatter",
+  "pylint.enabled": true,
+  "pylint.args": [
+    "--load-plugins",
+    "pylint_django"
+  ],
+}
```
Ishankoradia, line 8: why is this here ?

fatchat: added the `settings.json` to the repo because it seems useful to have here
added `pylint_django` to stop vscode from underlining our `Model.objects` everywhere

fatchat:
todo: send discord notifications to platform admins

.vscode/settings.json
Ishankoradia: It doesn not underline my. 
So i think we should remove this. This is a local dev environment setting.

ddpui/api/webhook_api.py
```diff
@@ -123,7 +126,11 @@ def post_notification_v1(request):  # pylint: disable=unused-argument
     if send_failure_notifications:
         org = get_org_from_flow_run(flow_run)
         if org:
-            email_flowrun_logs_to_orgusers(org, flow_run["id"])
+            email_flowrun_logs_to_superadmins(org, flow_run["id"])
+            notify_platform_admins(org, flow_run["id"])
```
Ishankoradia, line 130: For discord notifications to us (platform admins), can we add the org type whether its client or poc or demo or whatever the categories are. 

Also can we add the state whether its `failed` or `crashed`. Crashed should be a bigger concern for us.

ddpui/utils/webhook_helpers.py
```diff
@@ -105,9 +114,49 @@ def email_orgusers_ses_whitelisted(org: Org, email_body: str):
...
+    error, response = create_notification(
+        NotificationDataSchema(author="Dalgo", message=message, recipients=recipients)
+    )
+    if error:
+        logger.error(f"Error creating notification: {error}")
+        return
+    logger.info(f"Notification created: {response}")
+
+    if hasattr(org, "preferences"):
+        orgpreferences: OrgPreferences = org.preferences
```
Ishankoradia, line 142: Can we remove this ? ITs already being done in `create_notification()`

ddpui/core/notifications_service.py
```diff
@@ -126,11 +118,23 @@ def create_notification(
     if not notification:
         return {"message": "Failed to sent notification."}, None
 
+    last_recipient_id = None
     for recipient_id in recipients:
+        last_recipient_id = recipient_id
         error = handle_recipient(recipient_id, scheduled_time, notification)
```
Ishankoradia, line 124: We need to check for all receipeint's org. 
Send the discord notification to all org based on preferences. 
This function doesnt know that all receipients might be from the same org.

ddpui/api/webhook_api.py
```diff
@@ -123,7 +126,11 @@ def post_notification_v1(request):  # pylint: disable=unused-argument
...
+            notify_org_managers(
+                org, "A jobs has failed, please visit https://dashboard.dalgo.in/ for more details"
```
Ishankoradia, line 132: do you want to pick the webapp_url from config or `.env` ? Failure notifications on staging will then look weird while testing

.vscode/settings.json
fatchat: did you pip install the requirements? my reasoning was that someone who clones the project shouldn't see errors. 

the formatting settings (the "black" formatter) should also be in source control btw... if they're not could you tell me how to add them

Ishankoradia: i dont quite remember but it didn't give me that error for a very long time now

we have pre commit to handle the formatting locally and in our ci.

ddpui/api/webhook_api.py
fatchat: done

ddpui/utils/webhook_helpers.py
fatchat: done

ddpui/core/notifications_service.py
fatchat: done

ddpui/api/webhook_api.py
fatchat: done

Ishankoradia:
Screenshot 2024-12-05 at 10 57 46

Ran a sync on staging. This is the message i got for failure
Doesn't seem very useful. Should we atleast add the name of the pipeline or connection ? What do you think @fatchat

Ishankoradia (approved):

PR #910, opened 2024-12-05
for the time when generating a new orgdbt use the default value

Ishankoradia:


fatchat:
`gather_dbt_project_params` contains `dbt_binary = str(dbt_env_dir / "venv/bin/dbt")`

i think this is the last one

Ishankoradia:
We dont use that any more. We use `gather_dbt_project_params` via `DbtProjectManager.gather_dbt_project_params`.

If you see otherwise, I will change it. Have removed that funciton. Its not being used anywhere.

fatchat (approved):

PR #917, opened 2024-12-10
Fetching source schema catalog through celerytask

himanshudube97:


himanshudube97:
> _ No description provided. _

@fatchat Can you review it once. I tested it and it was working fine. I added a time.sleep(20) to the api request too to simulate a delay in the call. 
I need to check it for some error handling.

ddpui/websockets/airbyte_consumer.py
```diff
@@ -123,3 +129,96 @@ def websocket_receive(self, message):
...
+            self.respond(
+                WebsocketResponse(
+                    data={},
+                    message="Create an airbyte workspace first.",
+                    status=WebsocketResponseStatus.ERROR,
+                )
+            )
+            return
+
+        if "sourceId" not in payload:
```
fatchat, line 152: `or payload['sourceId'] is None`

ddpui/websockets/airbyte_consumer.py
```diff
@@ -123,3 +129,96 @@ def websocket_receive(self, message):
...
+            self.respond(
+                WebsocketResponse(
+                    data={},
+                    message="SourceId is required in the payload",
+                    status=WebsocketResponseStatus.ERROR,
+                )
+            )
+            return
+        source_id = payload.get("sourceId", None)
```
fatchat, line 161: then this can just become `source_id = payload['sourceId']`

ddpui/websockets/airbyte_consumer.py
```diff
@@ -123,3 +129,96 @@ def websocket_receive(self, message):
...
+
+        # creating a key and checking if it exists or not
+        task_key = f"{TaskProgressHashPrefix.SOURCE_SCHEMA_CATALOG}-{orguser.org.slug}-{source_id}"  # change this schema change to catalog
+        print(task_key, "taskkey ")
+        task_progress = SingleTaskProgress.fetch(task_key)
+
+        if task_progress is not None:
+            polling_celery(self, task_key)
```
fatchat, line 169: typically we will write this as `self.polling_celery(task_key)`

fatchat: (below as well)

ddpui/websockets/airbyte_consumer.py
```diff
@@ -123,3 +129,96 @@ def websocket_receive(self, message):
...
+            )
+        )
+    elif last_status == TaskProgressStatus.COMPLETED:
+        self.respond(
+            WebsocketResponse(
+                data={
+                    "status": last_status,
+                    "result": task_progress[-1]["result"],
+                },
+                message="Destination connection check completed",
```
fatchat, line 212: hmm?

ddpui/websockets/airbyte_consumer.py
himanshudube97: Added

ddpui/websockets/airbyte_consumer.py
```diff
@@ -123,3 +129,98 @@ def websocket_receive(self, message):
...
+        # creating a key and checking if it exists or not
+        task_key = f"{TaskProgressHashPrefix.SOURCE_SCHEMA_CATALOG}-{orguser.org.slug}-{source_id}"
+        task_progress = SingleTaskProgress.fetch(task_key)
+
+        if task_progress is not None:
+            polling_celery(consumer, task_key)
+            return
+
+        # This gives the task to celery
+        get_schema_catalog_task.delay(task_key, str(orguser.org.airbyte_workspace_id), source_id)
```
fatchat, line 173: when does this ever happen i.e. when would we find ourselves in this function without an STP

himanshudube97: So the flow is , 
1. Initially there will be NO STP, so the code will NOT go inside the 168-170 block. 
2. It will get to line 173 and create a STP in celery and then start the polling.

If I close the browser or refresh etc, and come back again to create connection and select the same source again.
1. The previous task would still be running in celery (if not completed)
2. The task_key would be same for the  same source.
4. So we will get a running STP, by the SingleTaskFunction.fetch(task_key) function line 166
3. Now the code block will enter 168-170 and start polling. (It will not give a new task to celery)

fatchat (approved):

PR #918, opened 2024-12-13
ability to call dbt cloud api as part of pipelines

himanshudube97:


fatchat (approved):
checked dbt cloud run as well as dbt core run

PR #928, opened 2025-01-13
Experiment/chat with data

Ishankoradia:
The way this would work.
1. Client establishes a connection by hitting the websocket endpoint `wss/chat/bot`
2. Then fetch the threads of the current orguser by sending message in the websocket `{action: "get_threads"}`
3. Get messages for the thread user selects by sending websocket message `{action: "get_messages", params: {thread_uuid: "", limit: 10, offset: 0}}`
4. Render the messages
5. User can now enter/ask the next question. Send it as a message in websocket `{action: "ask_bot", params: {message: "Get me top 5 students", thread_uuid: ""}}`. 
6. Since step 5 might take for the llm service to generate the response. Frontend will have to render it in the UI as message sent
7. Refresh the state and refetch the latest messages to show the bot's answer.

Messages over websocket will have to stringified


Ishankoradia:
While testing we should make sure we close the threads, and otherwise openai will be filled with assistants.

ddpui/websockets/chat_with_data_consumer.py
```diff
@@ -0,0 +1,260 @@
...
+        )
+
+    def close_thread(self, payload: CloseThreadRequest):
+        """
+        Close the thread
+        """
+        thread = Thread.objects.filter(uuid=payload.thread_uuid).first()
+        if thread:
+            llm_service.close_file_search_session(thread.session_id)
+            thread.delete()
```
himanshudube97, line 245: should we mark it as close ?? why are we deleting thread from db?

ddpui/websockets/chat_with_data_consumer.py
```diff
@@ -0,0 +1,300 @@
...
+        thread = Thread.objects.create(
+            uuid=uuid.uuid4(),
+            orguser=self.orguser,
+            session_id=session_id,
+            meta={"sql": payload.sql, **payload.meta},
+        )
+
+        self.respond(
+            WebsocketResponse(
+                data={"thread_uuid": str(thread.uuid)},
```
himanshudube97, line 155: here in the data = {}, we will should be returning the generated `sql query` too as we are displaying it in the frontend.

ddpui/websockets/chat_with_data_consumer.py
Ishankoradia: add a status column and will mark the thread as close but we will clear all open ai resources

(this will be used for the demo though)

fatchat:
Closing PR; keeping branch

PR #933, opened 2025-01-22
tasks in prefect deployment may not have the "connection_id" key

fatchat:
if the key is missing then we can't delete the connection


fatchat:
sorry that was for another issue 
have removed

Ishankoradia (approved):

PR #938, opened 2025-01-24
download preview data bug

Ishankoradia:
- `StringIO()` and `DictWriter` are being initialized multiple times , which is causing the data descrepancy while writing


fatchat:
tested from a juypter notebook and didn't see duplicates

fatchat (approved):

PR #939, opened 2025-01-29
added primaryKey field with cursor field

himanshudube97:


fatchat (approved):
well now we know why the cursor is an array

ddpui/ddpairbyte/airbyte_service.py
```diff
@@ -761,8 +761,10 @@ def create_connection(
             # weirdhly the cursor field is an array of single element eg ["created_on"] or []
             if schema_cat["config"]["syncMode"] == "incremental":
                 schema_cat["config"]["cursorField"] = [selected_streams[stream_name]["cursorField"]]
+                schema_cat["config"]["primaryKey"] = [selected_streams[stream_name]["primaryKey"]]
             else:
                 schema_cat["config"]["cursorField"] = []
+                schema_cat["config"]["primaryKey"] = []
```
fatchat, line 767: non-incremental should / might also have a primary key (in `selected_streams[stream_name]["primaryKey"]`)

ddpui/ddpairbyte/airbyte_service.py
```diff
@@ -815,8 +817,11 @@ def update_connection(
...
```
fatchat, line 823: similarly here

fatchat:
@himanshudube97 also tests have to go into `test_airbyte_service.py`

ddpui/ddpairbyte/airbyte_service.py
himanshudube97: I checked the airbyte ui and tried there. The only combination where i can add primary key is incremental + append/dedup. So in my opinion this is correct.

ddpui/ddpairbyte/airbyte_service.py
himanshudube97: Same as above: I checked the airbyte ui and tried there. The only combination where i can add primary key is incremental + append/dedup. So in my opinion this is correct.

Ishankoradia (approved):

PR #943, opened 2025-01-31
AIRBYTE_SOURCE_BLACKLIST

fatchat:
the marketplace commcare connector doesn't work the way we need it to, we want our clients to use our connector (until we submit it to the marketplace)

this PR blocks the airbyte marketplace version of commcare from showing up in the "new source" dropdown


ddpui/settings.py
```diff
@@ -248,6 +248,10 @@
     },
 }
 
+AIRBYTE_SOURCE_BLACKLIST = [
+    "airbyte/source-commcare",
+]
+
```
Ishankoradia, line 254: should we pick this from the env ?

Ishankoradia (approved):

ddpui/settings.py
fatchat: you're right, have made the change

PR #964, opened 2025-02-13
queueing workflows

Ishankoradia:


Ishankoradia:
queueing doesn't work as expected, maybe good to revisit next time. 

idea was to run cd workflow after ci on PR merge

PR #971, opened 2025-02-19
957 update the elementary profile when the dbt credentials is updated

fatchat:


fatchat:
this does not work because the `profiles/profiles.yml` doesn't get updated as soon as we update the `DbtCliProfile` block in prefect... we have to run a dbt operation first

Ishankoradia (approved):

PR #983, opened 2025-02-25
980 setup vanna training for orgs

Ishankoradia:
Steps to deplyoy this
1. Make sure you deploy the ai service changes
2. Update the `.env` with the template variables defined. 
3. Run the django command `scaffold-rag-training-on-warehouse.py` for any org(s). This command will 
     1. Create the vector db with the user and password
     2. Create the `OrgWarehouseRagTraining` config object for the corresponding warehouse
     4. Train the rag on warehouse schema based on step 2


ddpui/api/warehouse_api.py
```diff
@@ -366,7 +458,60 @@ def post_save_warehouse_prompt_session(
...
+        ).first()
+        # since its overwriting the old session, we need to keep/persist the orguser (created_by)
+        curr_session.orguser = old_session.orguser
+
+        if old_session:
+            old_session.delete()
+            logger.info(f"Deleted the old session llm analysis {payload.old_session_id}")
+
+        if not old_session:
+            if payload.sql:
```
fatchat, line 504: aren't we ensuring `old_session` by this point

ddpui/celeryworkers/tasks.py
```diff
@@ -862,37 +867,28 @@ def summarize_warehouse_results(
...
+            session_status=LlmSessionStatus.RUNNING,
+            session_type=LlmAssistantType.LONG_TEXT_SUMMARIZATION,
+            request_meta={"sql": sql},
         )
-        llm_session.session_status = LlmSessionStatus.FAILED
+    else:
+        llm_session = LlmSession.objects.filter(id=llmsession_pk).first()
+        if not llm_session:
+            raise Exception(f"Llm session with primary key {llmsession_pk} not found ")
+        llm_session.session_status = LlmSessionStatus.RUNNING
```
fatchat, line 883: what session statuses are expected here? i mean before we set it to `RUNNING` what could it be

Ishankoradia: you are right. anything but `RUNNING`

this is for the case when they summarize after the sql generation step. so just before it comes here, the status actually refers to the `sql generation` step.

Ishankoradia: pushed this change

ddpui/api/warehouse_api.py
Ishankoradia: `overwrite = True` has two meanings for different versions of the feature

the newer version can have `overwrite = True` & just update the sql inplace if the user decides to edit. for which the `old_session` can be none. another case in the same version, a user can also overwrite the session to save as new.

ddpui/celeryworkers/tasks.py
fatchat: ok so we expect to run it when the sql-generation step has already completed successfully

fatchat:
Closing the PR; keeping the branch open though

PR #986, opened 2025-03-03
Wren ai for clients

himanshudube97:


fatchat:
the url field shouldn't be nullable (or even blankable), because why would we have a db entry without the url
can you also write the api tests

fatchat (approved):

PR #997, opened 2025-03-12
Cancel a queued manual job

himanshudube97:
## Summary by CodeRabbit

- **New Features**
  - Introduced functionality that lets users cancel queued manual jobs associated with flow runs, with improved validations for user permissions and job references.
  - Enhanced task state management through more structured state representations for clearer status indications.

- **Bug Fixes**
  - Adjusted error handling to ensure appropriate responses for missing organization details and invalid flow run IDs.

- **Tests**
  - Added comprehensive tests to verify successful cancellations and proper error handling for scenarios such as missing organization details, invalid job references, and access restrictions.


fatchat:
1. no tests
2. the api endpoint should be a `POST`

fatchat:
in fact coderabbit has a third suggestion as well please see its comments

fatchat:
@Ishankoradia is there a way we can verify that the passed `flow_run_id` actually belongs to the requestor's `Org`? do we have it stored anywhere e.g. in a `TaskLock` or a `ConnectionJob`

Ishankoradia:
> @Ishankoradia is there a way we can verify that the passed `flow_run_id` actually belongs to the requestor's `Org`? do we have it stored anywhere e.g. in a `TaskLock` or a `ConnectionJob`

If the flow_run has reached a terminal state then yes. We could use the `deployment_id` in `ddpui_prefectflowrun` & figure out the `ddpui_orgdataflowv1` & then the org. 

If the flow has not reached a terminal state (it might or might not have taken a lock, its there is a lock ; yes `TaskLock` also has a flow_run_id reference & corresponding dataflow), we could just fetch from prefect `get_flow_run()` and use the `deployment_id` to do the same as above.

fatchat:
thanks ishan

@himanshudube97 so you need to call prefect-proxy's `proxy/flow_runs/<flow_run_id>` api (we probably have a service function to do this) and get the `deployment_id` from the response, and then use that to query the `OrgDataFlowv1` table from which you will get the `Org`. make sure the requesting `OrgUser` is attached to this org before proceeding

fatchat (approved):

PR #998, opened 2025-03-12
Delete sources ui4t Phase 1

Ishankoradia:


ddpui/core/dbtautomation_service.py
```diff
@@ -255,12 +256,108 @@ def get_output_cols_for_operation(org_warehouse: OrgWarehouse, op_type: str, con
...
+    """
+    if orgdbt_model.type == OrgDbtModelType.MODEL:
+        raise ValueError("Cannot delete a model as a source")
+
+    # delete entry in sources.yml on disk; & recreate the sources.yml
+    delete_dbt_source_in_project(orgdbt_model)
+
+    if cascade:
+        # delete all children of this model (operations & models)
+        cascade_delete_org_dbt_model(orgdbt_model)
```
fatchat, line 335: i thought we weren't doing cascading deletes for sources? is this code for future use

ddpui/core/dbtautomation_service.py
```diff
@@ -255,12 +256,108 @@ def get_output_cols_for_operation(org_warehouse: OrgWarehouse, op_type: str, con
...
+    else:
+        orgdbt_model.delete()
+
+
+def cascade_delete_org_dbt_model(orgdbt_model: OrgDbtModel):
```
fatchat, line 340: we don't cal `delete_dbt_model_in_project` for the models being removed here?

ddpui/core/dbtautomation_service.py
Ishankoradia: yes its for future use. Currently the param will always be `cascade=False`

ddpui/core/dbtautomation_service.py
Ishankoradia: good catch !

Ishankoradia: updated it

ddpui/core/dbtautomation_service.py
```diff
@@ -255,12 +256,109 @@ def get_output_cols_for_operation(org_warehouse: OrgWarehouse, op_type: str, con
...
+    """
+    if orgdbt_model.type == OrgDbtModelType.SOURCE:
+        raise ValueError("Cannot delete a source as a model")
+
+    operations = OrgDbtOperation.objects.filter(dbtmodel=orgdbt_model).count()
+
+    if operations > 0:
+        orgdbt_model.under_construction = True
+        orgdbt_model.save()
+    else:
```
fatchat, line 304: so if we have operations but `cascade is False` then we will delete the SQL file on disk but not the `OrgDbt` entry?

the cascade logic is confusing me

ddpui/core/dbtautomation_service.py
fatchat: but the parameter comes from the request doesn't it

ddpui/core/dbtautomation_service.py
Ishankoradia: removed the cascade logic for now

fatchat (approved):

PR #1008, opened 2025-03-28
missing task id for schema catalog

Ishankoradia:
The issue was we were polling before the task progress object was created in redis. 

Also the celery task was returning an object `HttpError` which would break in backend , since we cannot return objects that are not serializable from celery tasks.

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced error messages to provide clearer, more descriptive notifications during task failures.
  - Updated failure responses to ensure errors include detailed context.

- **New Features**
  - Added status notifications to signal the start of background tasks and indicate ongoing polling activities.

- **Style**
  - Standardized message capitalization for consistent and professional logging throughout the application.

### DIFF
--- a/ddpui/websockets/airbyte_consumer.py
+++ b/ddpui/websockets/airbyte_consumer.py
@@ -166,12 +166,19 @@ class SchemaCatalogConsumer(BaseConsumer):
         task_progress = SingleTaskProgress.fetch(task_key)
 
         if task_progress is not None:
+            logger.info(
+                f"Looks like a schema change task {task_key} is already under way, lets poll it"
+            )
             polling_celery(self, task_key)
             return
 
         # This gives the task to celery
+        logger.info(f"Starting a new catalog read celery task {task_key}")
+        taskprogress = SingleTaskProgress(task_key, 600)
+        taskprogress.add(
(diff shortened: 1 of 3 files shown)


ddpui/websockets/airbyte_consumer.py
```diff
@@ -166,12 +166,19 @@ def websocket_receive(self, message):
...
         if task_progress is not None:
+            logger.info(
+                f"Looks like a schema change task {task_key} is already under way, lets poll it"
+            )
             polling_celery(self, task_key)
             return
 
         # This gives the task to celery
+        logger.info(f"Starting a new catalog read celery task {task_key}")
+        taskprogress = SingleTaskProgress(task_key, 600)
```
fatchat, line 177: don't you have to remove the corresponding line in `get_schema_catalog_task` now

fatchat (approved):
approving but please see comment above

ddpui/websockets/airbyte_consumer.py
Ishankoradia: no we dont have to remove. it needs to load the same object from redis

Ishankoradia:
> approving but please see comment above

Resolved over a call

PR #1019, opened 2025-04-07
cms datainsights issue

Ishankoradia:
Initially we were constructing a connection string & passing it to the sqlalchemy engine

```
connection_string = (
            "postgresql://{encoded_username}:{encoded_password}@{host}/{database}".format(**creds)
        )
```

This required, the password and the username to be encoded.

Now we just pass the connection args dictionary which doesn't require encoding of parameters. 

```
connection_args = {
            "host": creds["host"],
            "port": creds["port"],
            "dbname": creds["database"],
            "user": creds["username"],
            "password": creds["password"],
        }
```

### DIFF
--- a/ddpui/datainsights/warehouse/postgres.py
+++ b/ddpui/datainsights/warehouse/postgres.py
@@ -24,8 +24,8 @@ class PostgresClient(Warehouse):
             "host": creds["host"],
             "port": creds["port"],
             "dbname": creds["database"],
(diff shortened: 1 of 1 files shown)


Ishankoradia:
Made this change on production & tested for cms. It works.

Ishankoradia:
`Git lens` vs code extension was super helpful in this debugging.

fatchat (approved):

PR #1030, opened 2025-04-17
make sure the lock objects read flow run status from our db

Ishankoradia:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved accuracy of lock status determination for tasks and pipelines by refining how flow run states are checked.
- **Chores**
	- Updated internal logic to directly use database models for flow run status checks, enhancing reliability.
	- Adjusted periodic task scheduling to conditionally skip a dataflow run time computation based on environment settings.
- **Tests**
	- Refined tests to use actual database records for flow runs instead of mocking external service responses.


fatchat:
Looks good, shouldn't we ensure that the lock has a flow run id though? Because it might not right

fatchat (approved):

PR #1038, opened 2025-04-23
Enhance Airbyte API with Pydantic models for source definitions and w…

Ashutoshx7:
## Summary by CodeRabbit

- **Refactor**
  - Improved API response consistency and validation by introducing standardized response formats for source definitions, workspace details, and connection information.
  - Simplified connection details returned to users, focusing on essential fields only.


ddpui/api/airbyte_api.py
```diff
@@ -66,7 +73,7 @@ def get_airbyte_source_definitions(request):
             if source_def["name"] in allowed_sources.split("|")
         ]
     logger.debug(res)
-    return res["sourceDefinitions"]
+    return AirbyteSourceDefinitionResponse(sourceDefinitions=res["sourceDefinitions"])
```
Ishankoradia, line 76: You will need to look at the fields in `sourceDefinitions` and then define your schema. It has more fields that just `sourceId` & `name`. Hit the api & see what you get, based on which you can define your schema. 

Also it would be nice to use ninja schema (which is built on top of pydantic) instead of BaseModel, just to keep things consistent.

ddpui/ddpairbyte/airbyte_service.py
```diff
@@ -126,7 +127,13 @@ def get_workspaces():
     return res
 
 
-def get_workspace(workspace_id: str) -> dict:
+# Example Pydantic schema for workspace response
+class WorkspaceResponse(BaseModel):
+    workspaceId: str
+    name: str
```
Ishankoradia, line 133: Same here, you should hit airbyte's api or look at the docs to see what a workspace api returns. It has more fields than these two

Ishankoradia (changes requested):

Ishankoradia:
Closing due to inactivity. Please feel free to reopen if you are going to work on it.

PR #1053, opened 2025-05-06
get airbyte and prefect urls from .env for notification emails

fatchat:


fatchat:
@himanshudube97 thank you!

Ishankoradia (approved):

PR #1076, opened 2025-06-11
1069 translate prefect deployment names for notification messages

fatchat:


ddpui/ddpairbyte/airbytehelpers.py
```diff
@@ -277,6 +278,10 @@ def create_connection(org: Org, payload: AirbyteConnectionCreate):
             sync_dataflow.clear_conn_dataflow = clear_dataflow
             sync_dataflow.save()
 
+            ConnectionMeta.objects.create(
```
Ishankoradia, line 281: what about when you update a connection ?

fatchat: good catch thanks
i've added and `update` inside `update_connection`
i didn't do `update_or_create` because we'll run the script when we release and going forward will assume that all connections have a connection-meta

PR #1079, opened 2025-06-11
Auth cleanup

Ishankoradia:
A pre-cursor refactor for the jwt PR (next).

This moves the auth middleware used by ninja to the top level object instead of using at a million places. 

Also cleanups the auth methods that we no longer use. 

## Summary by CodeRabbit

- **Refactor**
	- Simplified API authentication by removing explicit authentication middleware from individual route decorators and configuring global authentication middleware for all API routes.
	- Removed legacy and redundant authentication classes and functions to streamline authentication logic.
- **Chores**
	- Updated settings to remove explicit Django REST framework authentication class configuration.
- **Tests**
	- Removed tests related to deprecated authentication classes and functions.


Ishankoradia:
@fatchat this is ready to go in.

PR #1080, opened 2025-06-12
1075 implement jwt for dalgo auth

Ishankoradia:
## Summary by CodeRabbit

- **New Features**
  - Introduced JWT-based authentication, replacing the previous token-based system.
  - Added endpoints for login, logout with token blacklisting, and refreshing JWT access tokens.
  - Permissions are now managed via Redis and referenced in JWT tokens for improved security.
  - Added a periodic task to clean up expired blacklisted tokens daily.

- **Chores**
  - Updated environment variable templates and settings to support JWT configuration.
  - Added a new dependency for JWT support.
  - Updated CI workflows to include necessary secret environment variables.
  - Removed deprecated notification API endpoints.
  - Removed role fields from user and invitation models and related schemas.
  - Updated API authentication middleware to use the new JWT system.
  - Simplified organization user role handling and removed ownership transfer functionality.
  - Removed organization-specific Airbyte server overrides.
  - Adjusted imports and internal references to reflect role-related removals.
  - Cleaned up and updated tests to align with the new role model and JWT authentication.


fatchat:
the function `CustomTokenObtainSerializer.get_token` is only called from within `CustomTokenRefreshSerializer.validate` 

and this is the only time that the permissions are set in redis under `jwt_permissions:{uuid.uuid4()}`

once they expire we raise an exception in `CustomJwtAuthMiddleware.authenticate` 

if the user faces that exception, how does the flow get around to `get_token` again to recreate the permissions key

ddpui/auth.py
```diff
@@ -68,3 +73,85 @@ def authenticate(self, request, token):
...
+
+class CustomTokenObtainSerializer(TokenObtainPairSerializer):
+    @classmethod
+    def get_token(cls, user):
+        token = super().get_token(user)
+        # Generate permissions and store in Redis, put only the key in the token
+        orguser_permissions = {}  # { orguser_id : list[permission_slugs] }
+        for orguser in OrgUser.objects.filter(user=user):
+            role = orguser.new_role
+            orguser_permissions[orguser.id] = list(
```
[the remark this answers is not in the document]
Ishankoradia: by blow up , do you mean it will become very large if that user has lot of orgusers ?

ddpui/auth.py
```diff
@@ -64,7 +72,111 @@ def authenticate(self, request, token):
...
+                redis_client = RedisClient.get_instance()
+
+                orguser_role_map = redis_client.get(orguser_role_key)
+                orguser_role_map_json = json.loads(orguser_role_map)
+                # only fetch role if the orguser is not found in cache;
+                # for cases when a new orguser was added but the user is still logged in
+                if not orguser_role_map_json:
+                    orguser_role_id = orguser.new_role.id
+
+                orguser_role_id = orguser_role_map_json.get(str(orguser.id), None)
```
fatchat, line 118: `orguser_role_map_json` could be None

ddpui/auth.py
```diff
@@ -64,7 +72,111 @@ def authenticate(self, request, token):
...
+
+                permissions_map = redis_client.get(role_permissions_key)
+                permissions_json = json.loads(permissions_map)
+                if not permissions_json:
+                    raise HttpError(403, "Permissions not found or expired")
```
fatchat, line 123: instead of raising couldn't we look up the permissions from the `orguser.new_role`

ddpui/auth.py
```diff
@@ -64,7 +72,111 @@ def authenticate(self, request, token):
...
+        for role_perm in RolePermission.objects.select_related("permission").all():
+            if role_perm.role_id not in role_permissions:
+                role_permissions[role_perm.role_id] = []
+            role_permissions[role_perm.role_id].append(role_perm.permission.slug)
+
+        role_permissions_key = f"jwt_role_permissions:{uuid.uuid4()}"
+        redis_client = RedisClient.get_instance()
+        redis_client.setex(
+            role_permissions_key,
+            60 * 60 * 4,  # 4 hours expiry (should be >= access token lifetime)
```
fatchat, line 151: `role_permissions` is essentially a representation of the `RolePermission` table, there is only one. so we shouldn't make copies of it using these different uuid-based keys. let's just have one key and hardcode its name in `.env`. then we can take it out of the token as well

it also doesn't need to expire - when we reseed the table we should regenerate its value under the same key

Ishankoradia: Agreed.

ddpui/auth.py
```diff
@@ -64,7 +72,111 @@ def authenticate(self, request, token):
...
+
+        user_id = token_payload.get("user_id")
+        role_permissions_key = token_payload.get("role_permissions_key")
+        orguser_role_key = token_payload.get("orguser_role_key")
+
+        if not role_permissions_key:
+            raise HttpError(403, "Permissions key not found in token")
+
+        if token_payload and user_id:
+            request.user = User.objects.filter(id=user_id).first()
```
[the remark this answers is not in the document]
Ishankoradia: this we do but the one below i think can go away.

if a user doesn't have orguser mapped, they should still be able to login but not see anything. although this case might be very rare.

ddpui/auth.py
Ishankoradia: hmm yes we definitely could. makes sense.

ddpui/auth.py
```diff
@@ -64,7 +72,111 @@ def authenticate(self, request, token):
...
+
+        role_permissions_key = f"jwt_role_permissions:{uuid.uuid4()}"
+        redis_client = RedisClient.get_instance()
+        redis_client.setex(
+            role_permissions_key,
+            60 * 60 * 4,  # 4 hours expiry (should be >= access token lifetime)
+            json.dumps(role_permissions),
+        )
+
+        orguser_role_key = f"jwt_orguser_role:{uuid.uuid4()}"
```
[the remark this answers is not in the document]
Ishankoradia: Done

ddpui/auth.py
Ishankoradia: done

ddpui/auth.py
```diff
@@ -64,7 +72,111 @@ def authenticate(self, request, token):
...
+                q_orguser = q_orguser.filter(org__slug=orgslug)
+            orguser = q_orguser.select_related("org", "user").first()
+            if orguser is not None:
+                if orguser.org is None:
+                    raise HttpError(400, "register an organization first")
+
+                redis_client = RedisClient.get_instance()
+
+                orguser_role_map = redis_client.get(orguser_role_key)
+                orguser_role_map_json = json.loads(orguser_role_map)
```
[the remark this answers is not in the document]
Ishankoradia: handled the null checks

ddpui/auth.py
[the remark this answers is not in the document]
Ishankoradia: made this change

fatchat (approved):

PR #1084, opened 2025-06-23
prefect airbyte tasks timeout

Ishankoradia:


fatchat:
hmm but the thing is we need a script, not only now but anytime in future when we actually need to modify this setting across our deployments' config-tasks

fatchat (approved):

PR #1093, opened 2025-07-09
delete connections whose airbyte status is "deprecated"

fatchat:
celery task to delete a list of connections

cleanup: new dbthelpers.py module with create_or_update_org_cli_block from airbytehelpers

### DIFF
--- a/ddpui/celeryworkers/tasks.py
+++ b/ddpui/celeryworkers/tasks.py
@@ -7,10 +7,10 @@ from datetime import datetime, timedelta
 from subprocess import CompletedProcess
 import pytz
 from django.core.management import call_command
+from django.utils.text import slugify
 
 import yaml
 from celery.schedules import crontab
-from django.utils.text import slugify
 from ddpui.auth import ACCOUNT_MANAGER_ROLE
 from ddpui.celery import app, Celery
 
@@ -56,6 +56,7 @@ from ddpui.utils.taskprogress import TaskProgress
 from ddpui.utils.singletaskprogress import SingleTaskProgress
 from ddpui.utils.constants import (
     TASK_AIRBYTESYNC,
+    AIRBYTE_CONNECTION_DEPRECATED,
 )
 from ddpui.core.orgdbt_manager import DbtProjectManager
 from ddpui.ddpdbt.schema import DbtProjectParams
@@ -73,6 +74,7 @@ from ddpui.ddpprefect import DBTCLIPROFILE
 from ddpui.datainsights.warehouse.warehouse_factory import WarehouseFactory
 from ddpui.core import llm_service
 from ddpui.utils.helpers import (
+    find_key_in_dictionary,
     convert_sqlalchemy_rows_to_csv_string,
 )
 
@@ -417,6 +419,27 @@ def run_dbt_commands(self, org_id: int, task_id: str, dbt_run_params: dict = Non
         task_lock.delete()
 
 
+@app.task(bind=False)
+def delete_airbyte_connections(task_key: str, org_id, connection_ids: list[str]):
+    """deletes a set of airbyte connections and related dalgo artifacts"""
+    org = Org.objects.get(id=org_id)
+    # the task_key is keyed by org only
+    taskprogress = SingleTaskProgress(task_key, 180)
+    taskprogress.add({"message": "started", "status": TaskProgressStatus.RUNNING, "result": None})
+    for connection_id in connection_ids:
+        airbytehelpers.delete_connection(org, connection_id)
+        taskprogress.add(
+            {
+                "message": f"connection {connection_id} deleted",
+                "status": TaskProgressStatus.RUNNING,
+                "result": {},
+            }
+        )
+    taskprogress.add(
+        {"message": "all connections deleted", "status": TaskProgressStatus.COMPLETED, "result": {}}
+    )
+
+
 def detect_schema_changes_for_org(org: Org):
     """detect schema changes for all connections of this org"""
     org_tasks = OrgTask.objects.filter(org=org, task__slug=TASK_AIRBYTESYNC)
@@ -428,6 +451,8 @@ def detect_schema_changes_for_org(org: Org):
     ):
         schema_change.delete()
 
+    deprecated_org_tasks: list[OrgTask] = []
+
     # check for schema changes
     for org_task in org_tasks:
         connection_catalog, err = airbytehelpers.fetch_and_update_org_schema_changes(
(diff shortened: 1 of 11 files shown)


ddpui/celeryworkers/tasks.py
```diff
@@ -469,6 +483,14 @@ def detect_schema_changes_for_org(org: Org):
             except Exception as err:
                 logger.error(err)
 
+    for deprecated_org_task in deprecated_org_tasks:
+        logger.info(
+            f"deleting OrgSchemaChange for deprecated connection {deprecated_org_task.connection_id} for {org.slug}"
+        )
+        OrgSchemaChange.objects.filter(
+            org=org, connection_id=deprecated_org_task.connection_id
+        ).delete()
```
Ishankoradia, line 492: we can do this one trip to db. Fetch using `connection_id__in`

ddpui/celeryworkers/tasks.py
```diff
@@ -469,6 +483,14 @@ def detect_schema_changes_for_org(org: Org):
             except Exception as err:
                 logger.error(err)
 
+    for deprecated_org_task in deprecated_org_tasks:
```
Ishankoradia, line 486: what if the execution doesn't reach here and there is an error in some iteration of the above loop. The dangling schema change objects will still not be deleted.

ddpui/celeryworkers/airbytehelperasks.py
```diff
@@ -0,0 +1,37 @@
+from ddpui.celery import app
```
Ishankoradia, line 1: Can we put this function in `tasks.py` because there are whole lot of `airbytehelper` related tasks there.
If we are making a file separately for these tasks, then we will need to refactor `tasks.oy`

ddpui/ddpairbyte/deleteconnection.py
```diff
@@ -0,0 +1,93 @@
...
+    # delete all deployments
+    for dataflow in dataflows_to_delete:
+        logger.info("deleting prefect deployment %s", dataflow.deployment_name)
+        prefect_service.delete_deployment_by_id(dataflow.deployment_id)
+
+    # delete all dataflows
+    logger.info("deleting org dataflows from db")
+    for dataflow in dataflows_to_delete:
+        # if there is a reset and a sync then the dataflow will appear twice in this list
+        if dataflow.id:
```
Ishankoradia, line 45: why this check ? `id` will always be there, its a primary key right

ddpui/ddpairbyte/deleteconnection.py
```diff
@@ -0,0 +1,93 @@
...
+                else:
+                    tasks_to_keep.append(task)
+            parameters["config"]["tasks"] = tasks_to_keep
+            # logger.info(parameters)
+            if len(parameters["config"]["tasks"]) > 0:
+                payload = PrefectDataFlowUpdateSchema3(deployment_params=parameters)
+                prefect_service.update_dataflow_v1(dataflow_orgtask.dataflow.deployment_id, payload)
+                logger.info("updated deployment %s", dataflow_orgtask.dataflow.deployment_name)
+            else:
+                prefect_service.delete_deployment_by_id(dataflow_orgtask.dataflow.deployment_id)
```
Ishankoradia, line 79: you dont want to delete the dataflow from the db ?

ddpui/ddpairbyte/deleteconnection.py
```diff
@@ -0,0 +1,93 @@
+from ddpui.models.org import (
```
Ishankoradia, line 1: can we move this function to `airbytehelpers.py`. i dont see why we need a separate module(file) for just one function

and yes i know `airbytehelpers.py` has grown a lot but we can refactor that into meaningful modules later.

ddpui/ddpairbyte/deleteconnection.py
fatchat: oh right that check was from when the same object might have been in the list twice, i'll remove it

ddpui/celeryworkers/tasks.py
```diff
@@ -428,16 +430,28 @@ def detect_schema_changes_for_org(org: Org):
...
+            if os.getenv("ADMIN_EMAIL"):
+                send_text_message(
+                    os.getenv("ADMIN_EMAIL"),
+                    f"Schema change detection errors for {org.slug}",
+                    err,
+                )
             logger.error(err)
             continue
 
+        if connection_catalog["status"] == AIRBYTE_CONNECTION_DEPRECATED:
```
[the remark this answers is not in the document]
fatchat: ok i will send us an email and skip to the next connection

ddpui/celeryworkers/tasks.py
fatchat: yes... but the above loop just calls `fetch_and_update_org_schema_changes` and sends emails

ddpui/ddpairbyte/deleteconnection.py
fatchat: hmm yes if the prefect deployment goes then we should delete the dataflow as well

i found `orgtaskfunctions.delete_orgtask`, but that seems to have a bug it deletes all dataflows which contain the orgtask. shouldn't it check the config[tasks] like this function does?

ddpui/ddpairbyte/deleteconnection.py
fatchat: circular dependency...

ddpui/ddpairbyte/deleteconnection.py
fatchat: btw we have `on delete cascade` set for `OrgTask`s inside the `DataflowOrgTask` mapping object, isn't this wrong? when a `DataflowOrgTask` is deleted we shouldn't delete the `OrgTask` right?

fatchat: or does it mean that if the `OrgTask` is deleted then we should delete this mapping object

ddpui/celeryworkers/airbytehelperasks.py
fatchat: `tasks` imports `airbytehelpers`
and `airbytehelpers` has three functions which call this `delete_airbyte_connections` function
but `airbytehelpers` cannot import `tasks` because we will get a circular import, so i moved this function from `tasks` to `airbytehelpertasks`

ddpui/ddpairbyte/deleteconnection.py
fatchat: `airbytehelpers`  has three functions which call `delete_airbyte_connections` which calls `deleteconnection.deleteconnection`

so i can't put `deleteconnection` into `airbytehelpers`, we will get a circular dependency

Ishankoradia (approved):

Issue #873, opened 2024-10-08
Select column names for a connection - backend


fatchat:
https://docs.google.com/document/d/1OAkwfMIuLY3QJANJtTVJk_V2_bQL1H9HYRoHLdKiBkM/edit

![Image](https://github.com/user-attachments/assets/3403d33b-7c1b-4267-a918-3bcc662c33ac)

PR #1118, opened 2025-07-21
Select columns for stream - backend

MohitAgrawal16:
## Summary

Fixes: #873 
Added the logic of sending the columns data correctly in the way that airbyte required.

## Summary by CodeRabbit

* **New Features**
  * Improved stream configuration to support field-level selection, allowing users to choose specific columns for synchronization when setting up connections.
* **Tests**
  * Added comprehensive tests covering various scenarios of column selection and field-level sync configuration for streams.
* **Chores**
  * Updated `.gitignore` to exclude the `dbt-venv/` directory from version control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ __pycache__/
 # C extensions
 *.so
 
+dbt-venv/
 # Distribution / packaging
 .Python
 build/
--- a/ddpui/ddpairbyte/airbyte_service.py
+++ b/ddpui/ddpairbyte/airbyte_service.py
@@ -780,6 +780,34 @@ def create_connection(
                 schema_cat["config"]["cursorField"] = []
                 schema_cat["config"]["primaryKey"] = []
 
+            if (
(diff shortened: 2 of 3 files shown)


MohitAgrawal16:
@himanshudube97, what is the need of adding code to create_connection?

himanshudube97:
> @himanshudube97, what is the need of adding code to create_connection?

When the user will create a connection, then also he can select which stream to sync and which columns from the streams to sync right ?

fatchat:
tests are passing
observation: when the sync mode is changed from incremental to overwrite, the cursor selection should be voided

MohitAgrawal16:
@fatchat, for the primary key, when incremental is off, currently, we are not able to edit the primary key; does this mean we also need to clear the primary key when incremental is off?

fatchat:
@MohitAgrawal16 
hmm i don't know why incremental is related to primary-key selection...
do it the way you are suggesting i.e. primary-key only for incremental AND append/dedup

PR #1132, opened 2025-07-31
Fix -  Generating access token from refresh token.

himanshudube97:
## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured that access tokens generated during token refresh correctly include custom claims.

* **Documentation**
  * Clarified comments regarding token types and custom claims.
  * Fixed a typo in the class documentation for token refresh functionality.

### DIFF
--- a/ddpui/auth.py
+++ b/ddpui/auth.py
@@ -177,7 +177,7 @@ class CustomTokenObtainSerializer(TokenObtainPairSerializer):
 
     @classmethod
     def get_token(cls, user):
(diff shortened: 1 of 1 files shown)


himanshudube97:
In Django REST Framework SimpleJWT:

  - TokenObtainPairSerializer.get_token(user) returns a `RefreshToken` object
  - The `RefreshToken object has an .access_token `property that gives you the corresponding `AccessToken`

  So when we call:
`  token = super().get_token(user)  # This returns a RefreshToken`

Ishankoradia (approved):

PR #1269, opened 2026-03-19
Feature/report comments

siddhant3030:
## Summary by CodeRabbit

* **New Features**
  * Commenting on report snapshots and charts (create, edit, delete, list), @mentions with in-app notifications and optional emails, and "mark as read" actions with per-target read states.

* **API**
  * New Comments API endpoints and mentionable-users endpoint.

* **Migrations**
  * Comment and CommentReadStatus models with soft-delete and indexes.

* **Documentation**
  * Clarified package initializer guidance to avoid re-exports.

* **Tests**
  * Extensive tests covering comments, mentions, notifications, and read-state behavior.

* **Chores**
  * Added HTML email sender, mention email templates, and share-token helper for exports.


ddpui/core/comments/comment_service.py
```diff
@@ -0,0 +1,361 @@
...
+from ddpui.utils.custom_logger import CustomLogger
+
+from .exceptions import (
+    CommentNotFoundError,
+    CommentValidationError,
+    CommentPermissionError,
+)
+from .mention_service import MentionService
+
+logger = CustomLogger("ddpui.core.comments")
```
Ishankoradia: you dont need to pass the path here. just `comments` is fine. 
the logger already has the module file & line no. information

admin.py
siddhant3030: not sure why this file is here. will see and remove it

ddpui/schemas/comment_schema.py
```diff
@@ -0,0 +1,151 @@
...
+        emails = comment.mentioned_emails or []
+        mentions = []
+        if emails:
+            users_map = getattr(comment, "_mentioned_users_map", None)
+            if users_map is None:
+                users_map = {
+                    ou.user.email: ou
+                    for ou in OrgUser.objects.filter(user__email__in=emails).select_related("user")
+                }
+            mentions = [
```
Ishankoradia: why do we need to this ? for each ?

this is quite costly. we are already storing `mentioned_emails` in the comment table.

Did i miss something ?

Ishankoradia (changes requested):

ddpui/core/comments/comment_service.py
```diff
@@ -0,0 +1,361 @@
...
+
```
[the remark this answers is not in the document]
siddhant3030: yes we can do that. will do in the end

Ishankoradia (approved):

PR #1274, opened 2026-03-24
Remote infra git clone

Ishankoradia:
## Summary by CodeRabbit

* **New Features**
  * Multi-dbt Docker image for Prefect job runners with Prefect integrations and multiple dbt runtimes.
  * Added git-clone system task and pipeline support.

* **Improvements**
  * Centralized pipeline orchestration for consistent runs, clearer error mapping, schedule/run enhancements, and single-task/manual DBT run support.
  * New optional exclude_git parameter for transformation task listing.

* **Documentation**
  * Added README with build and deployment guidance for the multi-dbt image.

* **Config**
  * Queue config and migration flow now include an EKS workpool flag (is_workpool_eks).


ddpui/management/commands/migrate_org_queue.py
```diff
@@ -58,13 +62,19 @@ def add_arguments(self, parser):
...
+            help="Mark the workpool as EKS workpool (default: False)",
+        )
 
     def handle(self, *args, **options):
         org_slug: str = options["org_slug"]
         queue_type: str = options["queue_type"]
         new_queue: str = options["new_queue"]
         new_workpool: str = options.get("new_workpool")
         dry_run: bool = options["dry_run"]
+        is_workpool_eks: bool = options["is_workpool_eks"]
```
pradnk, line 77: @Ishankoradia How would we use this option? If this is false then where does the work pool go? Can we leverage this to bring up infra for local vs distrubuted perhaps?

Ishankoradia: Yes correct, thats what the flag is for.

If its false, things stay as is and pipelines run from a local process workpool (same as current state).

if we switch that then, the pipelijes will start runngin on eks

ddpui/core/orchestrate/pipeline_service.py
```diff
@@ -0,0 +1,795 @@
...
+        logger.info(f"Pipeline has {len(sync_orgtasks)} airbyte syncs")
+
+        # Handle transform tasks
+        dbt_project_params = None
+        dbt_orgtasks = []
+        git_orgtasks = []
+        dbt_cloud_orgtasks = []
+        orgdbt = org.dbt
+
+        if payload.transformTasks and len(payload.transformTasks) > 0:
```
himanshudube97: i see the similar logic in update pipeline also . can we make a common function out of this ?

Ishankoradia: Good call. Let me refactor.

PR #1283, opened 2026-03-27
feat: add multi-tab support to dashboards

NaveenCode:
**Summary**

Added tabs JSONField to the Dashboard model to store per-tab layout and components
Created DashboardTabSchema Pydantic schema to validate tab structure (id, title, layout_config, components)
Updated DashboardUpdate schema to accept an optional tabs list for saving tab state
Updated DashboardResponse schema to return tabs in API responses
DashboardService.create_dashboard() now initializes every new dashboard with a default tab ("Untitled Tab 1") with empty layout and components
DashboardService.update_dashboard() handles saving tabs by converting Pydantic models to dicts before JSON storage
Updated dashboard copy (duplicate) logic in native API to deep-copy tabs and remap filter IDs in both layout_config and components of each tab so copied dashboards have correct references

** Tasks Completed **

Create a new dashboard — verify it is saved with a single default tab in the DB (tabs field)
Update a dashboard with multiple tabs — verify all tab data (title, layout, components) is persisted correctly
Copy/duplicate a dashboard with tabs and filters — verify filter IDs in tab layout and components are remapped to the new filter IDs
Fetch a dashboard via GET — verify tabs array is returned in the response
Existing dashboards without tabs field load correctly (empty tabs: [] returned)
Verify no regression on existing layout/components/filter endpoints

## Summary by CodeRabbit

* **New Features**
  * Dashboards now support multiple tabs; each tab stores its own layout and components and is returned in API responses.
  * Dashboard creation and updates handle tabs (new dashboards get a default tab).
  * Dashboard duplication preserves and remaps per-tab state and filter associations.

* **Chores**
  * Migration and a management command (with a --dry-run option) to migrate existing dashboards to the tabs format.

* **Tests**
  * Added tests covering tab creation, updates, duplication, and filter remapping.


ddpui/services/dashboard_service.py
```diff
@@ -280,10 +281,19 @@ def create_dashboard(data: DashboardData, orguser: OrgUser) -> Dashboard:
         Returns:
             Created Dashboard instance
         """
+        # Generate default tab for new dashboard
+        default_tab = {
+            "id": f"tab-{int(time.time() * 1000)}",
```
siddhant3030: do we really need this? dashboard id will be different right?

NaveenCode: Yes, the Dashboard id is different from the tab id. 
Dashboard id identifies the dashboard, and Tab id identifies 
  which tab within that dashboard. A dashboard can have multiple tabs, so we need
   a way to reference each one.

ddpui/api/dashboard_native_api.py
```diff
@@ -242,9 +242,43 @@ def duplicate_dashboard(request, dashboard_id: int):
 
             updated_components[new_component_id] = new_component_data
 
-        # Update the dashboard with the corrected layout_config and components
+        # Copy tabs and update filter IDs inside each tab
```
NaveenCode: Lines 245-276 handle copying tabs with filter ID updates — this is NOT dead    
  code, it's the actual working code after migration since all data will live in 
  tabs.                                                                          
                                                                                 
  You're right that some existing code will become dead after migration:         
  - Lines 210-243 and 279-280 handle root-level layout_config and components     
  - After migration, these fields will always be empty [] and {}                 
  - This code existed before this PR for backward compatibility during the       
  transition                                                                     
                                                                                 
  Once migration is confirmed on all environments, we can clean up the root-level
   handling in a follow-up PR and make tabs the single source of truth.

siddhant3030 (commented):
reviewing

ddpui/api/dashboard_native_api.py
siddhant3030: do we really need this after the migration?

ddpui/schemas/dashboard_schema.py
```diff
@@ -22,6 +22,19 @@ class DashboardCreate(Schema):
     grid_columns: int = 12
 
 
+class DashboardTabSchema(Schema):
+    """Schema for a dashboard tab
+
+    Each tab contains its own layout and components, allowing users
+    to organize charts into separate views within a single dashboard.
+    """
```
siddhant3030, line 30: `title: str` has no length constraint, but the comment says "max 50 chars" and the frontend enforces 50. Add `title: str = Field(..., max_length=50)` so the backend validates too. Same for `id` — consider a min_length or regex to prevent empty strings.

siddhant3030 (commented):
Three architecture guideline notes — see inline comments.

siddhant3030:
**Architecture note — missing feature-specific exceptions**

The dashboard module has no `core/dashboards/exceptions.py`. Tab validation errors (e.g. invalid tab ID, empty title) would surface as generic Python errors. Per our guidelines, each feature should define its own exception hierarchy (`DashboardNotFoundError`, `DashboardValidationError`, etc.) so the API layer can map them to proper HTTP status codes. Not a blocker for this PR since it's pre-existing, but worth tracking as tech debt.

ddpui/api/dashboard_native_api.py
NaveenCode: When we copy a dashboard, we also need to copy the tabs inside it. Since tabs are stored as a JSON field on the dashboard row (not a separate table), they don't get carried over automatically — we have to explicitly copy them and remap any filter IDs they reference to the new filter IDs created for the duplicated dashboard. Migration has nothing to do with this.

ddpui/schemas/dashboard_schema.py
NaveenCode: you're right on both points.

title in DashboardTabSchema now has Field(..., max_length=50) to match the frontend constraint and the comment. id has Field(..., min_length=1) to prevent empty strings from passing validation.

ddpui/schemas/dashboard_schema.py
```diff
@@ -31,6 +44,7 @@ class DashboardUpdate(Schema):
     target_screen_size: Optional[str] = None
     layout_config: Optional[list[dict]] = None
```
Ishankoradia: shouldn't we stop taking in `layout_config` now, since all information is in `tabs` ? (atleast for update)

And we are doing a data migration when we release this.

ddpui/services/dashboard_service.py
```diff
@@ -360,10 +361,19 @@ def create_dashboard(data: DashboardData, orguser: OrgUser) -> Dashboard:
         Returns:
             Created Dashboard instance
         """
+        # Generate default tab for new dashboard
+        default_tab = {
```
Ishankoradia: Should we use the schema `DashboardTabSchema` here to validate the structure  ?

ddpui/services/dashboard_service.py
```diff
@@ -414,6 +424,9 @@ def update_dashboard(
             dashboard.layout_config = data.layout_config
         if data.components is not None:
             dashboard.components = data.components
+        if data.tabs is not None:
+            # Convert Pydantic models to dicts for JSON storage
+            dashboard.tabs = [tab.dict() for tab in data.tabs]
```
Ishankoradia: We should use `tab.model_dump()` since we have moved to pydantic v2

ddpui/management/commands/migrate_dashboards_to_tabs.py
```diff
@@ -0,0 +1,95 @@
...
+                    "id": f"tab-{int(time.time() * 1000)}",
+                    "title": "Untitled Tab 1",
+                    "layout_config": dashboard.layout_config or [],
+                    "components": dashboard.components or {},
+                }
+
+                # Update dashboard structure
+                dashboard.tabs = [default_tab]
+                dashboard.layout_config = []
+                dashboard.components = {}
```
Ishankoradia: i would leave the `layout_config` & `components` as is & not set it to empty states. If something goes we'll be able to rollback.

Once we are done with the release and the migration, we should remove these columns from the database in a follow up PR and do a cleanup. 

Let me know what you think.

ddpui/management/commands/migrate_report_snapshots_to_tabs.py
```diff
@@ -0,0 +1,106 @@
...
+            else:
+                default_tab = {
+                    "id": f"tab-{int(time.time() * 1000)}",
+                    "title": "Untitled Tab 1",
+                    "layout_config": layout_config,
+                    "components": components,
+                }
+
+                frozen["tabs"] = [default_tab]
+                frozen["layout_config"] = []
```
Ishankoradia: Same thing here. Dont make this empty for now.

We can submit a follow up PR to cleanup these columns from both dashboards and reports.

ddpui/services/dashboard_service.py
NaveenCode: right ,updated

ddpui/services/dashboard_service.py
NaveenCode: Yes, done. We now use DashboardTabSchema(...).model_dump() instead of a raw dict so Pydantic validates the structure before saving.

ddpui/management/commands/migrate_dashboards_to_tabs.py
NaveenCode: Agreed. We'll leave layout_config and components as-is during migration — only writing tabs. This keeps the original data intact for rollback if anything goes wrong post-release. Once the release is stable, we'll remove those columns from the DB in a follow-up cleanup PR.

code updated

ddpui/management/commands/migrate_report_snapshots_to_tabs.py
NaveenCode: Agreed , changes done

ddpui/schemas/dashboard_schema.py
NaveenCode: Agreed. Removed layout_config and components from DashboardUpdate — all canvas data now flows through tabs only. Updated the service to stop applying root-level fields on update as well. The migration handles existing dashboards. Frontend's save payload is also cleaned up to only send tabs.

siddhant3030 (approved):

PR #1311, opened 2026-04-21
Update filtering logic to handle `read_status=0` correctly using `read_status is not None`

### LINKED ISSUE #1309: Notification history read_status filter is not applied correctly
## Summary

The `/notifications/history` endpoint accepts a `read_status` query parameter, but filtering does not appear to work as expected.

## Findings

In `ddpui/api/notifications_api.py`, the route accepts `read_status` but passes `None` to the downstream function:

```python
notifications_functions.get_notification_history(
    page, limit, read_status=None
)
```

Additionally, in `ddpui/core/notifications/notifications_functions.py`, filtering currently uses:

```python
if read_status:
```

which may skip filtering when `read_status=0`.

## Impact

Requests such as:

```text
/notifications/history?read_status=1
```

or

```text
/notifications/history?read_status=0
```

may not return properly filtered results.

## Suggested Direction

* Forward the incoming `read_status` parameter from the API layer
* Consider checking `read_status is not None` in the downstream function

dipak0000812:
## Summary

Fixes #1309 by correctly applying the `read_status` query parameter in `/notifications/history`.

## Changes Made

* Forward incoming `read_status` from the API handler to `get_notification_history()`
* Update filtering logic to use `read_status is not None`

## Why

Previously:

* API route always passed `None`, so filtering was ignored
* `if read_status:` skipped filtering when `read_status=0`

## Result

* `read_status=1` returns read notifications
* `read_status=0` returns unread notifications
* no parameter returns all notifications

## Summary by CodeRabbit

* **Bug Fixes**
  * Request validation added for the read/unread filter: only 0 or 1 are accepted; invalid values return a 400 error.
  * Notification history filtering now respects falsey values (e.g., unread), and returns de-duplicated results so users see accurate, non-repeated notification lists.

### DIFF
--- a/ddpui/api/notifications_api.py
+++ b/ddpui/api/notifications_api.py
@@ -46,7 +46,7 @@ def get_notification_history(request, page: int = 1, limit: int = 10, read_statu
     Returns all the notifications including the
     past and the future scheduled notifications
     """
-    error, result = notifications_functions.get_notification_history(page, limit, read_status=None)
+    error, result = notifications_functions.get_notification_history(page, limit, read_status=read_status)
     if error is not None:
         raise HttpError(400, error)
 
--- a/ddpui/api/notifications_api.py
+++ b/ddpui/api/notifications_api.py
@@ -46,6 +46,9 @@ def get_notification_history(request, page: int = 1, limit: int = 10, read_statu
     Returns all the notifications including the
     past and the future scheduled notifications
     """
+    if read_status is not None and read_status not in (0, 1):
+        raise HttpError(400, "read_status must be 0 or 1")
+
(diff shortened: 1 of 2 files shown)


dipak0000812:
Thanks for the review. I rechecked the models and updated the query to filter through the reverse relation from `Notification` to `NotificationRecipient` using `notifications_received__read_status`, with `distinct()` to avoid duplicate notifications when multiple recipients match.

dipak0000812:
Thanks for the suggestion. I’ve added validation in the API layer so only `0` and `1` are accepted for `read_status`, while omitting the parameter still returns all notifications.

PR #1321, opened 2026-04-29
last modified added

siddhant3030:
## Summary by CodeRabbit

* **New Features**
  * Snapshots now record and display who last modified the summary (shows editor's email in snapshot metadata); snapshot view always includes a last-modified field (null if never edited).
* **Database**
  * Migration added to persist the last-modified user on snapshots.
* **Tests**
  * Extended test coverage for last-modified behavior, summary update flows, and view data.


ddpui/api/report_api.py
```diff
@@ -155,7 +155,7 @@ def update_snapshot(request, snapshot_id: int, payload: SnapshotUpdate):
     """Update a snapshot"""
     orguser: OrgUser = request.orguser
     try:
-        snapshot = ReportService.update_snapshot(snapshot_id, orguser.org, payload)
+        snapshot = ReportService.update_snapshot(snapshot_id, orguser.org, payload, orguser)
```
himanshudube97: We can just pass orguser from here and inside the function can access org = orguser.org , feels like duplication.

ddpui/core/reports/report_service.py
```diff
@@ -428,7 +428,7 @@ def list_snapshots(
             query &= Q(created_by__user__email__icontains=created_by_email)
         return list(
             ReportSnapshot.objects.filter(query)
-            .select_related("created_by__user")
+            .select_related("created_by__user", "last_modified_by__user")
```
himanshudube97: list reports page doesn't show last_modifed_by field. we can remove it from here.

ddpui/models/report.py
```diff
@@ -73,6 +73,14 @@ class ReportSnapshot(models.Model):
         null=True,
         help_text="User who created this snapshot",
     )
+    last_modified_by = models.ForeignKey(
```
himanshudube97, line 80: can we rename it to something like executive_summary_last_modified_by -> you can make it shorter, but looking at the model it looks like this is the last modified for reports. thoughts ?

siddhant3030: The intent is for this field to track any user-initiated modification to the snapshot not just summary edits. So last_modified_by is intentionally generic. we dont know what yet

ddpui/api/report_api.py
siddhant3030: done

ddpui/core/reports/report_service.py
siddhant3030: done, good catch

ddpui/models/report.py
himanshudube97: cool.

PR #1323, opened 2026-04-30
handle ssl cert in dbt cli config

Ishankoradia:
## Summary by CodeRabbit

* **Bug Fixes**
  * SSL handling: map provided SSL "mode" to the DBT profile's sslmode.
  * SSL CA certificates are kept as in-memory content and no longer written to disk.
  * Validation added: a project directory is required when supplying a CA certificate.

* **Tests**
  * Added tests covering sslmode-only behavior, CA certificate handling without disk writes, and missing project-directory error.

### DIFF
--- a/ddpui/core/dbtfunctions.py
+++ b/ddpui/core/dbtfunctions.py
@@ -41,10 +41,15 @@ def map_airbyte_destination_spec_to_dbtcli_profile(
         if mode:
             conn_info["sslmode"] = mode
 
-        if ca_certificate and dbt_project_params.org_project_dir:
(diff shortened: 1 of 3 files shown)


siddhant3030 (commented):
**ddpui/core/dbtfunctions.py:48** — `sslrootcert_content` isn't consumed anywhere in DDP_backend. Is there a paired prefect-proxy PR that reads this key and writes the cert to disk before dbt runs? Without it, `sslrootcert` will point to a file that doesn't exist.

siddhant3030 (approved):

PR #1396, opened 2026-06-01
fix: treat equals on timestamp columns as full day range to match tim…

NaveenCode:
## Summary
- Fixed date filter not working when filtering on timestamp columns (e.g. `_airbyte_extracted_at`)
- When a user picks a date like `2026-02-10`, it now correctly returns all rows from that full day instead of returning no results

## Summary by CodeRabbit

* **New Features**
  * Date-only filtering (yyyy-MM-dd) for timestamp/timestamptz/datetime now treats equals/not_equals as whole-day ranges and shifts greater/less-than boundaries to the next-day where appropriate.
  * Timestamp date-equals filters are excluded from grouped multi-equals handling to preserve day-boundary semantics.

* **Tests**
  * Added tests for equals/not_equals day inclusion/exclusion, boundary shifts for greater/less operators, grouping behavior, and null handling across timestamp types.


ddpui/core/charts/charts_service.py
```diff
@@ -654,14 +678,26 @@ def apply_chart_filters(
...
+            query_builder.where_clause(
+                and_(column(column_name) >= value, column(column_name) < _next_day(value))
+            )
+        elif operator == "not_equals" and _is_timestamp_date(filter_config):
+            query_builder.where_clause(
+                or_(column(column_name) < value, column(column_name) >= _next_day(value))
+            )
+        elif operator == "greater_than":
+            v = _next_day(value) if _is_timestamp_date(filter_config) else value
+            query_builder.where_clause(column(column_name) >= v)
```
himanshudube97: if the column is a non datetime field, will this work ? 
you are doing column_name >= v and its a greater_than operator

NaveenCode: Oh, yes, you are right, for non-datetime fields it should be > only not = >, I merged both, sorry fixing

PR #1406, opened 2026-06-10
Revert "feat: add org logo upload and branding support"

NaveenCode:
- Add logo_url, logo_s3_key, logo_filename fields to Org model (migration 0164)
- Add org logo service with upload from file, upload from URL, and delete logic
- Add org logo API endpoints: GET, POST /upload/, POST /url/, DELETE
- Return logo_filename in OrgSchema and OrgLogoResponse for frontend tab state
- Fix SVG logo blurriness in PDF by converting to high-res PNG via canvas before capture

## Summary by CodeRabbit

* **New Features**
  * Added org logo management (upload from file or URL, retrieve via image proxy, and delete)
  * Public dashboards and report views now expose org logo URL
  * Improved PDF export handling for external SVG images (more reliable rasterization)
* **Chores**
  * Added `S3_IMAGES_BUCKET` placeholder for shared image storage
* **Tests**
  * Added API and service tests for org logo success/error paths, plus S3 utility coverage


himanshudube97 (approved):

ddpui/utils/s3_utils.py
```diff
@@ -0,0 +1,90 @@
...
+        Body=file_bytes,
+        ContentType=content_type,
+    )
+
+    logo_url = _build_s3_url(bucket, region, s3_key)
+    logger.info(f"Uploaded org logo for {org_slug} to s3://{bucket}/{s3_key}")
+    return logo_url, s3_key
+
+
+def delete_org_logo(s3_key: str) -> None:
```
Ishankoradia: can make you this like a generic s3 util for uploading files so it can be used by other services in the future

we should also refactor elementary_service to use your new utility since that also has some s3 fucntions being called.

ddpui/core/org_logo/org_logo_service.py
```diff
@@ -0,0 +1,152 @@
...
+            raise OrgLogoNotFoundError()
+        return org
+
+    @staticmethod
+    def get_logo_bytes(org: Org) -> tuple[bytes, str]:
+        """Fetch raw logo bytes from the stored URL for server-side proxying."""
+        if not org.logo_url:
+            raise OrgLogoNotFoundError()
+        try:
+            resp = http_requests.get(org.logo_url, timeout=10, stream=True)
```
Ishankoradia: we have an http client. can't we use that ?

you can update it, if some params are missing

ddpui/core/org_logo/org_logo_service.py
```diff
@@ -0,0 +1,152 @@
...
+    OrgLogoS3Error,
+    OrgLogoFetchError,
+)
+
+logger = CustomLogger("ddpui.core.org_logo")
+
+
+class OrgLogoService:
+    @staticmethod
+    def get_logo(org: Org) -> Org:
```
Ishankoradia: this function doesn't make sense

its called get_logo but it returns`Org`.

also this feels over refactoring - that condition can just be checked in the api layer. we dont make functions for business logic

ddpui/core/org_logo/org_logo_service.py
```diff
@@ -0,0 +1,152 @@
...
+    OrgLogoNotFoundError,
+    OrgLogoValidationError,
+    OrgLogoS3Error,
+    OrgLogoFetchError,
+)
+
+logger = CustomLogger("ddpui.core.org_logo")
+
+
+class OrgLogoService:
```
Ishankoradia: do we need a separate service for  just uploading logo ?

i feel its an overkill. we should merge this with user_org_api. and maybe use `orgfunctions`

let me know what you think

ddpui/api/org_logo_api.py
```diff
@@ -0,0 +1,116 @@
...
+from ddpui.schemas.org_schema import OrgLogoResponse, OrgLogoUrlPayload
+
+logger = CustomLogger("ddpui.org_logo_api")
+
+org_logo_router = Router()
+
+
+@org_logo_router.get("/", response=ApiResponse[OrgLogoResponse])
+@has_permission(["can_view_orgusers"])
+def get_org_logo(request):
```
Ishankoradia: this api file is an overkill, just for uploading a logo which is related to org we are creating new file ?

cant we use `user_org_api.py`

.env.template
```diff
@@ -66,6 +66,7 @@ S3_AWS_ACCESS_KEY_ID=
 S3_AWS_SECRET_ACCESS_KEY=
 ELEMENTARY_S3_BUCKET=
 LATEST_ELEMENTARY_PACKAGE_VERSION=
+S3_IMAGES=                            # bucket for org images (logos, etc.)
```
Ishankoradia: can you make this a bit more clear. 

maybe S3_ORG_LOGO_BUCKET

ddpui/api/org_logo_api.py
```diff
@@ -0,0 +1,116 @@
...
+    """Get the current org logo"""
+    orguser: OrgUser = request.orguser
+
+    try:
```
Ishankoradia: why do we need a separate api to get the org logo url ?

can't we just store it in the orguser context when we fetch details of the current user

ddpui/core/org_logo/org_logo_service.py
```diff
@@ -0,0 +1,152 @@
...
+            Updated Org instance
+        """
+        # Delete old S3 file if the previous logo was uploaded (not a URL)
+        if org.logo_s3_key:
+            try:
+                delete_org_logo(org.logo_s3_key)
+            except Exception as e:
+                logger.warning(f"Failed to delete old S3 logo for {org.slug}: {e}")
+
+        org.logo_url = image_url
```
Ishankoradia: we need to add some validations to this url and make sure its an image. 

people might set random urls. what if its a script ?

ddpui/utils/s3_utils.py
NaveenCode: Done — s3_utils.py now has generic upload_file, delete_file, and download_file functions. Refactored elementary_service to use download_file instead of calling S3 directly.

ddpui/core/org_logo/org_logo_service.py
NaveenCode: updated, Used dalgo_get from ddpui/utils/http.py instead of importing requests directly. Added a raw=True param to dalgo_get so it returns the response object for binary/streaming use cases instead of always calling .json().

ddpui/core/org_logo/org_logo_service.py
NaveenCode: Done — removed get_logo from the service. The if not org.logo_url check is now inlined directly in the API endpoint, and the OrgLogoResponse is returned from there.

ddpui/core/org_logo/org_logo_service.py
NaveenCode: agree, but the logo operations (upload, delete, proxy) are distinct enough to live in their own service — same as any other feature in this codebase. Merging into user_org_api would make an already long file harder to navigate.

ddpui/api/org_logo_api.py
NaveenCode: Logo upload is a separate action from org creation — we set it up after the org exists, not during. user_org_api.py is also already very long. Keeping it in its own file makes both files cleaner and easier to navigate.

.env.template
NaveenCode: Updated S3_IMAGES to S3_IMAGES_BUCKET — this bucket is for all image uploads across the app, not just org logos, so the name should reflect that.

ddpui/api/org_logo_api.py
NaveenCode: Done. Logo URL is already in /api/currentuserv2, so the separate GET endpoint was redundant.

ddpui/core/org_logo/org_logo_service.py
NaveenCode: Added URL scheme validation (HTTP/HTTPS only) and a HEAD request check to verify the Content-Type is a valid image type before saving.

ddpui/api/org_logo_api.py
Ishankoradia: The user_org_api is not long, its just 730 lines.

A separate api file is usually for a core feature. and this is not that.

I still think we should move to the user_org_api.py 

i am not asking you to add logo while setting up the org, it can be a separate patch api (or maybe reuse some patch api for org if there exists).

ddpui/core/org_logo/org_logo_service.py
Ishankoradia: I disagree

its not about the distinctness of the api. It has to be a core feature or some external sevrice communication to call it a `service`. 

This is just a column update in org table (logo columns). 

we should move this to user_org_api. that file is not long, its just 730 lines. you can move the business logic to `orgfunctions`

ddpui/utils/s3_utils.py
Ishankoradia: have you tested this out locally and see if it works correctly, nothing breaks ?

ddpui/core/org_logo/org_logo_service.py
Ishankoradia: Have you tested it out by adding some url that is not an imaeg ?

ddpui/core/org_logo/org_logo_service.py
```diff
@@ -0,0 +1,189 @@
...
+def _get_logo_bucket() -> str:
+    bucket = os.getenv("S3_IMAGES_BUCKET")
+    if not bucket:
+        raise OrgLogoS3Error("S3_IMAGES_BUCKET environment variable is not set")
+    return bucket
+
+
+class OrgLogoService:
+    @staticmethod
+    def get_logo_bytes(org: Org) -> tuple[bytes, str]:
```
Ishankoradia: why are we passing the entire org object ? 

can't we just pass the logo_url: str -> this way the validation is done by pydantic & we dont need to `raise OrgLogoNotFoundError()`

does that make sense ?

pydantic(ninja) schemas are for this reason so that we dont have to write such stuff `if not org.logo_url`

ddpui/api/org_logo_api.py
```diff
@@ -0,0 +1,100 @@
...
+        return api_response(success=True, message="Logo deleted successfully")
+    except OrgLogoNotFoundError as e:
+        raise HttpError(404, str(e)) from e
+    except OrgLogoS3Error as e:
+        raise HttpError(502, str(e)) from e
+
+
+@org_logo_router.get("/proxy/")
+@has_permission(["can_view_orgusers"])
+def proxy_org_logo_image(request):
```
Ishankoradia: why do we need this api ? aren't we storing the logo on s3 and the s3 url is being embedded on the frontend.

am i missing something ?

ddpui/core/org_logo/org_logo_service.py
```diff
@@ -0,0 +1,189 @@
...
+            try:
+                delete_file(_get_logo_bucket(), org.logo_s3_key)
+            except Exception as e:
+                logger.error(f"S3 delete failed for {org.slug}: {e}")
+                raise OrgLogoS3Error("Failed to delete logo from S3") from e
+
+        org.logo_url = None
+        org.logo_s3_key = None
+        org.logo_filename = None
+        org.save(update_fields=["logo_url", "logo_s3_key", "logo_filename"])
```
Ishankoradia: what is difference between saving like this vs just doing `org.save()` ?

NaveenCode: update_fields only updates the specified columns in the DB, avoiding accidental overwrites of other fields changed concurrently. org.save() would update all fields.last

ddpui/core/org_logo/org_logo_service.py
NaveenCode: Done — changed get_logo_bytes to accept logo_url: str instead of the full Org object. The if not org.logo_url check is now inlined in the API layer before calling the service.

ddpui/api/org_logo_api.py
NaveenCode: The proxy exists for the chart download feature inside dashboards. When a user downloads a chart as PNG, the frontend calls /api/org/logo/proxy/ to fetch the logo and draws it onto an HTML Canvas to add branding. The browser enforces CORS when drawing an external-origin image (like an S3 URL) directly onto canvas — the canvas becomes "tainted" and canvas.toBlob() throws a security error, breaking the PNG download. The proxy sidesteps this by serving the image bytes from the same origin, so the canvas draw succeeds.

ddpui/core/org_logo/org_logo_service.py
NaveenCode: updated, as you suggested

ddpui/api/org_logo_api.py
NaveenCode: updated, as you suggested

ddpui/utils/s3_utils.py
NaveenCode: Yes, tested and working fine

ddpui/api/user_org_api.py
```diff
@@ -734,3 +742,79 @@ def get_iframe_token(request):
         "expires_in": 120,  # 2 minutes in seconds
         "org_slug": orguser.org.slug,
     }
+
+
+@user_org_router.post("/org/logo/upload/", response=ApiResponse[OrgLogoResponse])
```
Ishankoradia: for the naming of routes we usually follow REST api practices. Can you read about it a bit ?  and then rename the route accordingly. 

as much as possible we should follow that.

ddpui/api/user_org_api.py
```diff
@@ -734,3 +742,79 @@ def get_iframe_token(request):
...
+            message="Logo URL saved successfully",
+        )
+    except OrgLogoValidationError as e:
+        raise HttpError(400, str(e)) from e
+    except Exception as e:
+        logger.error(f"Failed to save logo URL for {orguser.org.slug}: {e}")
+        raise HttpError(500, "Failed to save logo URL") from e
+
+
+@user_org_router.delete("/org/logo/", response=ApiResponse)
```
Ishankoradia, line 795: here you have followed REST convention. nice !!

ddpui/api/user_org_api.py
```diff
@@ -734,3 +742,79 @@ def get_iframe_token(request):
...
+    try:
+        orgfunctions.delete_logo(orguser.org)
+        return api_response(success=True, message="Logo deleted successfully")
+    except OrgLogoNotFoundError as e:
+        raise HttpError(404, str(e)) from e
+    except OrgLogoS3Error as e:
+        raise HttpError(502, str(e)) from e
+
+
+@user_org_router.get("/org/logo/proxy/")
```
Ishankoradia: Same. Can you follow REST conventions in naming this route too?

We dont need the word `proxy`.

ddpui/tests/api_tests/test_org_logo_api.py
```diff
@@ -0,0 +1,137 @@
...
+from unittest.mock import patch, MagicMock
+
+import pytest
+from ninja.errors import HttpError
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "ddpui.settings")
+os.environ["DJANGO_ALLOW_ASYNC_UNSAFE"] = "true"
+django.setup()
+
+from django.contrib.auth.models import User
```
Ishankoradia: We should move these tests to user_org_api.py

We no longer have `org_logo_api.py`

ddpui/api/user_org_api.py
NaveenCode: updated, thanks

ddpui/api/user_org_api.py
NaveenCode: updated , thanks

ddpui/tests/api_tests/test_org_logo_api.py
NaveenCode: updated

Ishankoradia (approved):

PR #1409, opened 2026-06-11
delete chart config from dashboard , when we delete the chart

himanshudube97:
[https://linear.app/dalgo/issue/DALGO-1418/when-a-chart-is-deleted-it-leaves-this-empty-block-with-an-error](https://linear.app/dalgo/issue/DALGO-1418/when-a-chart-is-deleted-it-leaves-this-empty-block-with-an-error)

## Summary by CodeRabbit

* **Bug Fixes**
  * Charts are now automatically removed from all organization dashboards when deleted.
  * Chart deletion operations use atomic transactions to prevent orphaned chart tiles on dashboards.

* **Tests**
  * Added comprehensive tests for chart deletion cascade cleanup across dashboards.

### DIFF
--- a/ddpui/services/chart_service.py
+++ b/ddpui/services/chart_service.py
@@ -7,6 +7,7 @@ separating it from the API layer for better testability and maintainability.
 from typing import Dict, List, Optional, Tuple, Any
 from dataclasses import dataclass
 
+from django.db import transaction
 from django.db.models import Q
 
 from ddpui.models.visualization import Chart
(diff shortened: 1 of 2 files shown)


ddpui/services/chart_service.py
```diff
@@ -260,11 +261,59 @@ def delete_chart(chart_id: int, org: Org, orguser: OrgUser) -> bool:
             raise ChartPermissionError("You can only delete charts you created.")
 
         chart_title = chart.title
-        chart.delete()
```
Ishankoradia, line 263: @himanshudube97 do we really need to do this ?

We purposely didn't do it in the first go. It was because we didn't want the dashboard layout to look weird once they delete the chart (with a missing chart). We said (if i remember correctly), they can go and remove it from the dashboard first & then delete the chart.

Thats why there is a message saying which dashboard(s) the chart is part of

himanshudube97: oh ok. right. closing this pr.

PR #1421, opened 2026-07-07
 Implement audit log service for user actions

NaveenCode:
Summary
This PR adds backend audit logging for key user-initiated actions across authentication, organization, and workflow APIs. It introduces a shared audit log service, model, and migration, and records successful actions asynchronously.

Changes

Added audit log model, service, and database migration
Wired audit logging into auth and org-user flows
Added audit logging for pipeline, Airbyte, dbt, and transform actions
Added basic tests for audit log behavior

## Summary by CodeRabbit

* **New Features**
  * Added audit logging for key account and organization events (sign-in/out, password and email changes, invites, role updates, org logos, and user/org lifecycle actions).
  * Introduced a new cookie-based login token flow for embedded access.
* **Bug Fixes**
  * Improved consistency of password reset, email verification, and password change results.
  * Improved org-invitation resend handling when an invitation record is missing.
  * Enhanced role update logging to include field-level changes.


ddpui/core/audit_log_service.py
```diff
@@ -0,0 +1,186 @@
...
+        )
+    except Exception as err:
+        # If DB write fails, log it but don't crash
+        logger.error("audit_log_service: failed to write audit log", exc_info=err)
+    finally:
+        # Always close the connection to return it to the pool
+        django.db.connection.close()
+
+
+def compute_changes(
```
pradnk: @NaveenCode where is compute_changes called from? I see tests but not usage.

ddpui/api/user_org_api.py
pradnk: @NaveenCode Not seeing audit log call here. I see you calling it in post_forgot_password_v2 above in line 452 which is the request for password reset. But, this function is actually setting a new password so we will need an audit log here - which is more important to record change happened

ddpui/api/user_org_api.py
```diff
@@ -394,10 +450,24 @@ def post_forgot_password_v2(
...
+    if user:
+        orguser = OrgUser.objects.filter(user=user).first()
+        if orguser and orguser.org:
+            create_audit_log(
+                org=orguser.org,
+                orguser=orguser,
+                resource_type=AuditLogResourceType.AUTH,
+                resource_id="",
+                resource_name="password_reset_request",
+                action=AuditLogAction.UPDATE,
```
pradnk: @NaveenCode since this is a request for password reset I agree we should record it but is it UPDATE action as nothing is actually updated yet right? Thoughts?

ddpui/api/user_org_api.py
```diff
@@ -498,9 +567,41 @@ def post_organization_user_accept_invite_v1(
...
+            # Also log "user added to org"
+            create_audit_log(
+                org=invited_by_org,
+                orguser=new_orguser,
+                resource_type=AuditLogResourceType.ORG_USER,
+                resource_id=str(new_orguser.id),
+                resource_name=invited_email,
+                action=AuditLogAction.CREATE,
+            )
+
```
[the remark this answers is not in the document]
pradnk: @NaveenCode do check this

ddpui/api/user_org_api.py
```diff
@@ -517,31 +618,58 @@ def get_invitations_v1(request):
...
+        org=orguser.org,
+        orguser=orguser,
+        resource_type=AuditLogResourceType.INVITATION,
+        resource_id=str(invitation_id),
+        resource_name=invitation.invited_email if invitation else "",
+        action=AuditLogAction.UPDATE,
+        field_changes={"action": {"old": None, "new": "resent"}},
+    )
+
     return {"success": 1}
```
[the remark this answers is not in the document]
pradnk: Makes sense to do this too @NaveenCode

ddpui/api/user_org_api.py
NaveenCode: Ah, yes you are right, it should be a create not updated becuase it is creating a token so making it Create

ddpui/api/user_org_api.py
NaveenCode: yes now the update action should come there, updating

ddpui/core/audit_log_service.py
```diff
@@ -0,0 +1,186 @@
...
+            resource_type=AuditLogResourceType.DASHBOARD,
+            resource_id="123",
+            resource_name="Sales Dashboard",
+            action=AuditLogAction.DELETE,
+        )
+    """
+    try:
+        # Start a background thread to do the actual write
+        # daemon=True means the thread will be killed when main program exits
+        t = threading.Thread(
```
pradnk, line 77: @NaveenCode Does it make sense to use celery for this or threading is sufficient? I understand we are not having high traffic but celery can retry if connection to DB was cut for example - just more robust is the main point 

@Ishankoradia tagging to hear your input on this too

ddpui/api/user_org_api.py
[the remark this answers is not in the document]
NaveenCode: This scenario cannot occur in Dalgo's current flow. The invite_user_v1 function already checks if the user has membership in the org   and returns 'user already has an account' error. So when a user reaches accept_invitation, a new OrgUser is always created.  
  No additional check needed.

ddpui/api/user_org_api.py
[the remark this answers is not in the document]
NaveenCode: This is fixed, now if we will try to resend the invitation, and id didn't found then it will give an error like the invitation did not found

ddpui/core/audit_log_service.py
NaveenCode: used it with some endpoint like role update and invitation accepted and will be used more next milestones

ddpui/api/airbyte_api.py
```diff
@@ -127,6 +137,13 @@ def put_airbyte_source(request, source_id: str, payload: AirbyteSourceUpdate):
     if orguser.org.airbyte_workspace_id is None:
         raise HttpError(400, "create an airbyte workspace first")
 
+    # Capture state before update for field_changes tracking
+    try:
+        old_source = airbyte_service.get_source(orguser.org.airbyte_workspace_id, source_id)
+        old_name = old_source.get("name", "")
+    except Exception:
+        old_name = ""
```
Ishankoradia: Why do we need to do this ?

Cant we save the name of the current one ?

At 10:00am - created_source {name: "old_name"} resource_id = xyz

At 10:30am - update_source {name: "new_name"} resource_id = xyz

I can just run a query on the resource_id and see what changes were done.

ddpui/api/airbyte_api.py
```diff
@@ -114,6 +136,17 @@ def post_airbyte_source(request, payload: AirbyteSourceCreate):
         payload.config,
     )
     logger.info("created source having id " + source["sourceId"])
+    create_audit_log(
+        org=orguser.org,
+        orguser=orguser,
+        resource_type=AuditLogResourceType.DATA_SOURCE,
+        resource_id=source["sourceId"],
+        resource_name=source.get("name", payload.name or ""),
```
Ishankoradia: resource_name and resource_fields['name"] are redundant right ?

i think it should just stay in the `resource_fields`

ddpui/api/dbt_api.py
```diff
@@ -290,7 +346,18 @@ def post_dbt_makedocs(request):
...
-    return {"token": token.hex}
+    result = {"token": token.hex}
+
+    create_audit_log(
+        org=orguser.org,
+        orguser=orguser,
+        resource_type=AuditLogResourceType.DBT,
+        resource_id=str(orguser.org.dbt.id) if orguser.org.dbt else str(orguser.org.id),
+        action=AuditLogAction.EXECUTE,
+        resource_fields={"org": orguser.org.slug},
```
Ishankoradia: why are we adding org in json field when there is a foreign key of org_id in the table

ddpui/api/dbt_api.py
```diff
@@ -309,6 +376,15 @@ def put_dbt_schema_v1(request, payload: OrgDbtTarget):
     org.dbt.save()
     logger.info("updated orgdbt")
 
+    create_audit_log(
+        org=org,
+        orguser=orguser,
+        resource_type=AuditLogResourceType.DBT,
+        resource_id=str(org.dbt.id),
+        action=AuditLogAction.UPDATE,
+        resource_fields={"org": org.slug, "default_schema": payload.target_configs_schema},
```
Ishankoradia: same here, we dont need org key in the json payload.

ddpui/api/dbt_api.py
```diff
@@ -360,6 +436,15 @@ def post_run_dbt_commands(request, payload: TaskParameters = None):
 
     run_dbt_commands.delay(org.id, orgdbt.id, task_id, payload.model_dump() if payload else None)
 
+    create_audit_log(
+        org=org,
+        orguser=orguser,
+        resource_type=AuditLogResourceType.DBT,
+        resource_id=str(orgdbt.id),
+        action=AuditLogAction.EXECUTE,
+        resource_fields={"org": org.slug},
```
Ishankoradia: same here

ddpui/api/kpi_api.py
```diff
@@ -92,9 +95,25 @@ def get_kpi_summary(request):
...
+    org = orguser.org
 
     try:
         kpi = KPIService.create_kpi(payload, orguser)
+
+        resource_fields = payload.model_dump(exclude={"metric_id"})
+        resource_fields[
+            "name"
+        ] = kpi.name  # actual resolved name, not the raw (possibly None) payload value
+        resource_fields["metric"] = kpi.metric.name
```
Ishankoradia, line 106: we are doing this twice - fetching the metric. once here and the other time in the create_kpi() service fucntion.

why ?

ddpui/api/kpi_api.py
```diff
@@ -121,10 +140,33 @@ def get_kpi(request, kpi_id: int):
...
+        # what to touch. A field missing here means "not touched", not "cleared".
+        # "name" is the exception: always included (current value) so the row
+        # stays self-identifying without a separate name column.
+        touched = payload.model_dump(exclude_unset=True)
+        resource_fields = {k: v for k, v in touched.items() if k != "metric_id"}
+        if "metric_id" in touched:
+            resource_fields["metric"] = kpi.metric.name
+        resource_fields["name"] = kpi.name
+
+        create_audit_log(
```
Ishankoradia, line 161: same comment here. metric is being queried twice once here for audit and other in service function.

ddpui/api/kpi_api.py
```diff
@@ -138,16 +180,30 @@ def update_kpi(request, kpi_id: int, payload: KPIUpdate):
 def delete_kpi(request, kpi_id: int):
     """Delete a KPI"""
     orguser: OrgUser = request.orguser
+    org = orguser.org
+
+    # Capture KPI name before deletion
+    kpi = KPI.objects.filter(id=kpi_id, org=org).first()
+    kpi_name = kpi.name if kpi else str(kpi_id)
```
Ishankoradia: same here - kpi is being fetched twice.

ddpui/api/kpi_api.py
NaveenCode: this is the function in service file  at returning time - 
KPI.objects.select_related("metriccreated_byuser", "created_by__user").get(
            id=kpi.id
        )
so select_realted() populate all related foreign  keys and it loads the complete kpi object.So in kpi_api at  resource_fields["metric"] = kpi.metric.name it is not calling another query

ddpui/api/kpi_api.py
NaveenCode: true, fixed it so it would not refetch kpi again for audit log , used returned name from service file

ddpui/api/dbt_api.py
NaveenCode: True, fixed and fixed in other resources also ,thanks

Ishankoradia (approved):

PR #1425, opened 2026-07-12
Fix -> Calculated metric and saved metric were not applying to pie, number and map charts. 

himanshudube97:
## Summary by CodeRabbit

* **New Features**
  * Added support for expression-based metrics in number and pie charts.
  * Enhanced calculated metrics handling for map charts, including consistent metric naming in outputs.

* **Bug Fixes**
  * Corrected map data transformation to properly use metric expressions and aliases.
  * Improved count metric alias handling when no source column is provided.

* **Tests**
  * Added regression coverage for calculated metrics in map transforms.
  * Added SQL-generation assertions across chart types and metric kinds.


ddpui/core/charts/maps_service.py
```diff
@@ -177,6 +177,11 @@ def transform_data_for_map(
 
     customizations = customizations or {}
 
+    # Metrics may arrive as ChartMetric objects (payload.metrics) or dicts (resolved saved metrics).
+    # Normalize to dicts so the .get()-based access below works for both.
+    if metrics:
+        metrics = [m if isinstance(m, dict) else m.dict() for m in metrics]
```
Ishankoradia: @himanshudube97 why are we doing this ?

Cant we just type the input to `List[ChartMetric]`

Ishankoradia (approved):

ddpui/core/charts/maps_service.py
himanshudube97: updated it. made it consistent with other charts. thanks.

PR #1432, opened 2026-07-14
Admin Portal v1 (Week 1): platform-admin gate, org management, users tab (Issue #1254)

Veekshitha11:
Combined PR for the full Week 1 Admin Portal build (features/admin-portal/v1/plan.md), covering all four milestones: M1 (platform-admin gate), M2 (/admin shell + dashboard), M3 (organizations: list/create/edit), M4 (users tab: invite/role/remove/cancel-invite), plus a local-dev SES fallback and a full end-to-end lifecycle test. Org/user deactivation was scoped out of this PR -- the is_active enforcement added early on was removed rather than left half-wired, since no endpoint ever set it (see PR review discussion). Reintegrated onto current main -- originally built on an older base, then reconciled with two upstream changes that landed in the meantime: RBAC v2's role collapse (#1414, no behavior change to this work -- aliases preserved) and a cascade-delete fix (#1428) that independently fixed the same dashboard/chart-deletion bug this PR had flagged as a fast-follow; removal-impact messaging updated accordingly (orphaned, not deleted). Full regression suite green -- see commit messages for milestone-by-milestone detail.

## Summary by CodeRabbit

- **New Features**
  - Added a platform-admin portal for viewing platform health, statistics, organizations, and organization details.
  - Added organization creation and editing.
  - Added organization user management, including invitations, role changes, removal-impact previews, and membership deletion.
  - Added platform-admin status to current-user and login responses.
  - Added development email fallback logging when email credentials are unavailable.

- **Bug Fixes**
  - Improved invitation scoping and acceptance across organizations.
  - Authentication failures now return consistent HTTP 401 responses.


ddpui/api/admin_api.py
```diff
@@ -0,0 +1,325 @@
...
+
+def _admin_org_response(org: Org) -> AdminOrgSchema:
+    # the schema owns the field mapping (from_model); the API supplies the user_count,
+    # which needs a service call and so isn't the schema's to compute.
+    return AdminOrgSchema.from_model(org, admin_service.org_user_count(org))
+
+
+@admin_router.get("/ping", response=AdminPingSchema)
+@platform_admin_required
+def get_admin_ping(request):
```
siddhant3030: we dont need this. @Veekshitha11

ddpui/core/admin/admin_service.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+# --------------------------------------------------------------------------- #
+# Org create / update / (de)activate
+# --------------------------------------------------------------------------- #
+
+
+def create_org(payload: AdminCreateOrgSchema) -> Org:
+    """
+    Create an org and its plan. Takes the AdminCreateOrgSchema the API validated — the
```
siddhant3030: can we add only 2 lines of comments here? we dont need this long comment for every functions

ddpui/migrations/0182_invitation_invited_in_org.py
```diff
@@ -0,0 +1,46 @@
...
+# resolving to exactly the same org they did before. New admin-portal invites
+# set this explicitly to the target org (which may differ from invited_by.org
+# when a platform admin invites cross-org). accept/cancel read this field and
+# fall back to invited_by.org when null. See features/admin-portal/v1/plan.md §4.4.
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+def backfill_invited_in_org(apps, schema_editor):
```
siddhant3030, line 16: why do we need this?

ddpui/schemas/admin_schema.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+class AdminCurrentUserSchema(Schema):
+    """identity for the admin session — read by the frontend AdminGuard"""
+
+    email: str
+    is_platform_admin: bool
+
+
+class AdminPingSchema(Schema):
```
siddhant3030: we dont need this.

ddpui/schemas/admin_schema.py
```diff
@@ -0,0 +1,168 @@
...
+            id=org.id,
+            name=org.name,
+            slug=org.slug,
+            viz_url=org.viz_url,
+            base_plan=org.base_plan(),
+            user_count=user_count,
+        )
+
+
+class AdminCreateOrgSchema(Schema):
```
siddhant3030: is this extending current create org schema?

ddpui/utils/awsses.py
```diff
@@ -5,18 +5,43 @@
...
+
+logger = CustomLogger("ddpui")
 
 
 def _get_ses_client():
     """Get SES client instance - lazy initialization to avoid import-time failures"""
     return AWSClient.get_instance("ses")
 
 
+def _ses_available() -> bool:
```
siddhant3030: what is this?

siddhant3030: and why do we need it?

ddpui/utils/awsses.py
```diff
@@ -5,18 +5,43 @@
...
+    """True when real SES credentials are configured (the same pair AWSClient needs)."""
+    return bool(os.getenv("SES_ACCESS_KEY_ID") and os.getenv("SES_SECRET_ACCESS_KEY"))
+
+
 def send_text_message(to_email, subject, message):
```
siddhant3030, line 26: same question here?

siddhant3030:
**Review — two blocking items before merge:**

1. **Deactivation is enforceable but not actionable.** The PR description says M3/M4 include org/user deactivate, and the enforcement layer is all here (`Org.is_active`, `OrgUser.is_active`, migrations 0169/0170, the middleware 403s) — but no endpoint ever sets either flag, and neither `AdminOrgSchema` nor `AdminOrgUserSchema` exposes `is_active`, so the portal can't set *or display* active status (the `get_admin_org_users` docstring even promises "per-org Status"). If deactivation was deliberately deferred to a follow-up PR, please update the PR body and docstring; otherwise the endpoints are missing.

2. **Orphaned Airbyte workspace when plan creation fails.** In `post_admin_org`, `@transaction.atomic` rolls back the Org row when `create_org_plan` fails, but `create_organization` has already provisioned a real Airbyte workspace — an external side effect the DB rollback can't undo, leaving a leaked workspace with no Org pointing at it. Either delete the workspace in the plan-failure path, or log the workspace ID loudly for cleanup.

ddpui/migrations/0182_invitation_invited_in_org.py
Veekshitha11: Good question — this is needed because a platform admin invites cross-org: the admin isn't a member of the target org, so `invited_by.org` (the admin's own org) isn't the target org. Without `invited_in_org`, accept/cancel and the per-org pending-invites list would resolve to the wrong org. Trimmed the comment block down per your other note.

ddpui/utils/awsses.py
Veekshitha11: This is a local-dev convenience: `_ses_available()` checks whether real SES credentials are configured. When they're not (and DEBUG=True), `send_text_message` logs the email instead of trying to call real SES, so invite/signup/password-reset flows work on a laptop without needing AWS SES set up. It's off in staging/prod since DEBUG is off there, so it can't mask a real misconfiguration. There's a dedicated test for it. Shortened the docstrings per your other comment.

ddpui/utils/awsses.py
Veekshitha11: Same answer as above — `send_text_message` is the one function that uses this local-dev fallback (the other send_* helpers always hit real SES). Docstring trimmed.

ddpui/schemas/admin_schema.py
Veekshitha11: It wasn't before — now it does. Made AdminCreateOrgSchema(CreateOrgSchema) a real subclass overriding just base_plan/can_upgrade_plan/subscription_duration/superset_included with admin-friendly defaults, so admin_service.create_org can pass the payload straight through instead of manually rebuilding a CreateOrgSchema.

ddpui/api/admin_api.py
Veekshitha11: Removed /admin/ping and AdminPingSchema. It was only there to exercise the platform-admin gate in tests, so I repointed those 3 tests at /admin/currentuser (a real endpoint with the same gate) instead of dropping the coverage.

ddpui/core/admin/admin_service.py
Veekshitha11: Done — trimmed every function docstring in this file (and admin_api.py, which had the same style) down to 1-2 lines.

ddpui/schemas/admin_schema.py
Veekshitha11: Removed, along with the /admin/ping endpoint that used it.

Veekshitha11:
Addressed both:

1. **Deactivation** — removed `is_active` entirely rather than building the missing endpoint (migration `0172_remove_org_is_active_orguser_is_active`), since it wasn't part of what shipped. Also caught a stale `get_admin_org_users` docstring still promising "per-org Status" and fixed it. Updating the PR description to drop the stale deactivate mentions.
2. **Orphaned Airbyte workspace** — `admin_service.create_org` now calls `airbyte_service.delete_workspace(...)` when `create_org_plan` fails, before raising, so the workspace doesn't outlive the rolled-back Org row. A delete failure is logged loudly rather than silently swallowed.

Replied inline on the rest of the comments too.

ddpui/core/admin/admin_service.py
```diff
@@ -0,0 +1,365 @@
...
+    target_org_names = None
+    if notification.target_org_ids:
+        target_org_names = list(
+            Org.objects.filter(id__in=notification.target_org_ids).values_list("name", flat=True)
+        )
+    recipient_count = notification.notifications_received.count()
+    return AdminNotificationSchema.from_model(notification, target_org_names, recipient_count)
+
+
+def get_admin_notification_history() -> List[AdminNotificationSchema]:
```
himanshudube97: we will need to add pagination to the list history. 
For now simple ui can be -> have two tabs
a. create notification 
b. view history - the same table but with pagination. you can refrence our list tables - list chart, list dashbaord etc.

ddpui/api/admin_api.py
```diff
@@ -0,0 +1,360 @@
...
+
+
+@admin_router.post("/orgs", response=AdminOrgSchema)
+@platform_admin_required
+@transaction.atomic
+def post_admin_org(request, payload: AdminCreateOrgSchema):
+    """Create an org + plan. No OrgUser is attached — the first admin is invited on the
+    Users tab."""
+    try:
+        org = admin_service.create_org(payload)
```
himanshudube97: Ideally we won't create an org without a user. So instead of creating an empty org and then inviting admin from users tab, we should have a admin email section  in this screen. So now the org will be created and that x person (the member of the ngo who wants to join dalgo) will be the admin.

himanshudube97: @pradnk what do you think?

pradnk: Yes. Makes sense @himanshudube97 - earlier we used to be in the org so we could invite - let it not create an org without any users. Let it add 1 user who is admin

ddpui/api/admin_api.py
```diff
@@ -0,0 +1,369 @@
...
+
+logger = CustomLogger("ddpui")
+
+# No separate admin session: signs in through the normal POST /api/v2/login/, and
+# authority comes from @platform_admin_required on each route.
+admin_router = Router()
+
+
+@admin_router.get("/currentuser", response=AdminCurrentUserSchema)
+@platform_admin_required
```
Ishankoradia: @Veekshitha11 we are not using user attributes to identify super admin. 

We have a role , "Super admin" thats the platform admin. And ideally we should use the same decorator `has_permission` here also. You might need to seed some admin permisson for this role

Veekshitha11: Makes sense, will switch to the role. Dropping platform_admin_required, seeding a can_manage_platform permission for super-admin only, and gating the admin endpoints with @has_permission like everywhere else.

PR #1436, opened 2026-07-23
Feature/resource sharing v2

siddhant3030:
## Summary by CodeRabbit

* **New Features**
  * Added resource sharing for dashboards, charts, reports, and KPIs with view/edit access for people, groups, and pending invitations.
  * Added access requests, approvals, declines, privacy controls, and ownership transfers.
  * Added organization People and Groups management.
  * Added configurable access defaults, public-sharing settings, and per-resource access indicators.
  * Added access-aware viewing, editing, listing, duplication, locking, filtering, and sharing.
* **Bug Fixes**
  * Invitation acceptance now preserves group memberships and resource access.
  * Removed related shares and access requests when resources are deleted.


siddhant3030 (commented):
Three structural issues — all variants of the same "write it once" principle from the repo's own rules (layer architecture, empty __init__.py, service-delegation skill). Details inline.

ddpui/core/access/access_control.py
```diff
@@ -0,0 +1,238 @@
...
+        else:
+            granted = grants.get(str(resource.pk))
+            if resource.is_private:
+                levels[resource.pk] = _hide_no_access(granted)
+            else:
+                levels[resource.pk] = _hide_no_access(max_access_level(granted, floor))
+    return levels
+
+
+def get_access_map_for_resource(org, rtype: str, resource_id) -> dict[int, str]:
```
siddhant3030, line 146: Why can’t this decision logic be a single function? The same rule—owner/admin → edit; private → explicit grant only; otherwise max(grant, floor)—is duplicated three times in this file (`get_user_access`:93, `get_user_access_map`:121), and the `no_access` handling already differs.

Something like `effective_level(...)` would give us one place to read, change, and unit-test the rule, preventing the implementations from drifting.


```python
def effective_level(is_owner_or_admin, grant, floor, is_private):
    if is_owner_or_admin:
        return AccessLevel.EDIT
    if is_private:
        return grant or AccessLevel.NO_ACCESS
    return max_access_level(grant, floor) or AccessLevel.NO_ACCESS
```

ddpui/core/access/access_control.py
```diff
@@ -0,0 +1,238 @@
...
+- ``accessible_filter``    — the same rule as one ORM ``Q`` for list endpoints
+
+Precedence (first match wins):
+  1. creator or admin/super-admin        -> "edit" (full)
+  2. explicit USER grant                 -> its level ("no_access" -> None)
+  3. GROUP grants (max across groups)    -> best level
+  4. org-default floor                   -> default_analyst_level / default_member_level
+
+``no_access`` at step 2 is the explicit deny — the only way to revoke one
+principal below a permissive org floor. This is precedence, not max-merge,
```
siddhant3030, line 15: This docstring promise isn't what the code does: `_grants_map` ends with a max-merge of user and group levels, so an explicit USER `no_access` loses to any group grant (and to a permissive floor in `get_user_access`). Either make the code precedence-based as documented, or fix the docstring — right now the most-read lines of the file describe an algorithm the file doesn't implement.

ddpui/core/access/access_control.py
```diff
@@ -0,0 +1,238 @@
...
+
+def _org_floor(orguser: OrgUser) -> str:
+    """Org-default access level for this orguser's role. Missing row → model defaults."""
+    prefs = OrgPreferences.objects.filter(org=orguser.org).first() or OrgPreferences()
+    if orguser.new_role is not None and orguser.new_role.slug == RoleSlug.ANALYST:
+        return prefs.default_analyst_level
+    return prefs.default_member_level
+
+
+def _grants_map(
```
siddhant3030, line 51: nit: this map keys by `str(resource_id)`, so every caller writes `grants.get(str(...))` and `accessible_filter` converts back with `int(rid)`. Normalizing to one key type here would remove the juggling at all call sites.

ddpui/core/notifications/templates/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+- **Trial shell** (white header + navy wordmark, ``trial_shell._render_trial_email_shell``) —
+  verify / welcome / midpoint / pre-end / post-deletion / day-3 nudges.
+- **Biz-dev plain-text** (``biz_dev``) — internal team notifications, no shell.
+
+Each shell is defined exactly once; every ``render_*`` function delegates
+to the appropriate shell. ``test_shell_is_single_source_of_truth`` /
+``test_trial_shell_is_single_source_of_truth`` enforce the invariant.
+"""
+
+from ddpui.core.notifications.templates.alert import render_alert_email
```
siddhant3030, line 16: Repo invariant: core feature `__init__.py` stays empty (no re-exports, no `__all__`). This 58-line hub adds no behavior, puts an extra hop in every "where does this live" search, and re-exports two underscore-private names (`_render_email_shell`) as if they were public API. Suggest emptying it and importing from the concrete modules.

ddpui/core/access/access_control.py
```diff
@@ -0,0 +1,238 @@
...
+def accessible_filter(orguser: OrgUser, rtype: str) -> Q:
+    """The same rule as one ORM filter, for list endpoints: which resources of
+    this rtype may the orguser see at all (level >= view)? Compose it into the
+    list query with ``&`` — inaccessible rows never leave the database."""
+    if _is_admin(orguser):
+        return Q()
+
+    grants = _grants_map(orguser, rtype)
+    allowed_ids = [int(rid) for rid, level in grants.items() if level != AccessLevel.NO_ACCESS]
+    denied_ids = [int(rid) for rid, level in grants.items() if level == AccessLevel.NO_ACCESS]
```
siddhant3030, line 229: `resource_id` is a CharField with no numeric constraint, so one malformed row (`""`, `"12 "`, a future writer bug) makes `int(rid)` raise — and since this filter runs in every list endpoint, that one row 500s every list page for that user (or a whole group, if it's a group grant). `Q(id__in=...)` accepts strings directly, so the cast can just be dropped — or skip-and-log non-numeric rows.

PR #1437, opened 2026-07-24
Add per-user favorite support for charts and dashboards, so starring …

NaveenCode:
## Summary
- Add per-user favorite support for charts and dashboards — previously the star toggle only lived in frontend state and reset on every refresh
- New `ChartFavorite` / `DashboardFavorite` models (per-user, unique per chart/dashboard) with migration `0170_chart_dashboard_favorites`
- New endpoints: `POST`/`DELETE /api/charts/{id}/favorite/` and `POST`/`DELETE /api/dashboards/{id}/favorite/`
- `is_favorite` added to `ChartResponse`/`DashboardResponse`, computed only on `list_charts`/`list_dashboards` (the only consumers) — `get_chart`, `update_chart`, `get_dashboard`, `update_dashboard` intentionally don't compute it, keeping the change scoped to what's actually used
- Favorite state is scoped per user via `OrgUser`, verified against Superset's own `FavStar` model for consistency with how favoriting works elsewhere

## Summary by CodeRabbit

* **New Features**
  * Added the ability to favorite and unfavorite charts and dashboards.
  * Chart and dashboard listings now show each item’s favorite status.
  * Favorite actions are user-specific and safely handle repeated requests.
  * Missing charts or dashboards return a not-found response.
  * Favorites are automatically removed when their chart or dashboard is deleted.

* **Tests**
  * Added coverage for favorite status, user isolation, idempotency, deletion cleanup, and error handling.


ddpui/migrations/0170_chart_dashboard_favorites.py
```diff
@@ -0,0 +1,87 @@
+# Generated by Django 4.2 on 2026-07-24 07:50
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+    dependencies = [
```
Ishankoradia: @NaveenCode we dont need mutliple models for marking a resource favroutie.

checkout superset's `favstar` table in their meta database.

ddpui/models/favorite.py
```diff
@@ -0,0 +1,46 @@
...
+        return [(key.value, key.name) for key in cls]
+
+
+class Favorite(models.Model):
+    """Tracks which org users have favorited which resources. Favoriting is
+    personal — one user's favorite has no effect on any other user."""
+
+    org_user = models.ForeignKey(OrgUser, on_delete=models.CASCADE, related_name="favorites")
+    resource_type = models.CharField(max_length=20, choices=FavoriteResourceType.choices())
+    resource_id = models.BigIntegerField()
```
himanshudube97, line 19: If i delete a chart then it won't delete the rows in ths table right ? Same with the dashboard too. 
Is this the correct design choice ?

NaveenCode: Correct, it should be like what you are suggesting. Updated, thanks
Now, when deleting the chart/dashboard, entry for that chart/dashboard will also be removed from the favorite table

ddpui/services/chart_service.py
```diff
@@ -247,6 +249,7 @@ def delete_chart(chart_id: int, org: Org, orguser: OrgUser) -> str:
 
         chart_title = chart.title
         chart.delete()
+        FavoriteService.remove_favorites_for_resource(FavoriteResourceType.CHART, chart_id)
```
himanshudube97: what if chart delete function executes with success , but remove favorites fails ?

himanshudube97 (approved):

PR #1441, opened 2026-08-03
ingest and cast

Ishankoradia:
## Summary by CodeRabbit

* **New Features**
  * Airbyte connections now support optional post-sync transformations, including persistence during creation and updates.
  * Connection details display configured post-sync transformations.
  * Warehouse-level dbt profile configuration supports pipeline and EDR tasks.
  * Added tooling to copy dbt profile settings to warehouses, including filtering and dry-run support.

* **Bug Fixes**
  * Improved handling of missing warehouse configuration during deployment and cleanup.
  * Airbyte metadata updates continue gracefully when metadata persistence fails.
  * Unsupported warehouse types now return clear validation errors.

* **Changes**
  * Removed dbt Cloud task support from transformation workflows.


ddpui/ddpairbyte/airbytehelpers.py
```diff
@@ -285,6 +290,24 @@ def create_connection(org: Org, payload: AirbyteConnectionCreate):
                 connection_id=airbyte_conn["connectionId"], connection_name=payload.name
             )
 
+        # Always upsert the AirbyteConnection block so every connection has a corresponding
+        # Prefect block (carries the connection name + any post-sync ops). Consistent with
+        # update_connection which also always upserts.
+        try:
```
himanshudube97, line 355: If prefect upsert fails then we will be left with an airbyte connection and an cast op orgtask?

ddpui/ddpairbyte/airbytehelpers.py
```diff
@@ -646,6 +671,31 @@ def update_connection(org: Org, connection_id: str, payload: AirbyteConnectionUp
             connection_name=connection["name"]
         )
```
himanshudube97, line 731: this connectionMeta update and sync_org_task.save should also be wrapped inside a transaction right.

ddpui/ddpairbyte/airbytehelpers.py
```diff
@@ -646,6 +671,31 @@ def update_connection(org: Org, connection_id: str, payload: AirbyteConnectionUp
...
+    ).first()
+    if sync_org_task:
+        sync_org_task.post_sync_transform = payload.post_sync_transform or None
+        sync_org_task.save(update_fields=["post_sync_transform"])
+
+        server_block = OrgPrefectBlockv1.objects.filter(org=org, block_type=AIRBYTESERVER).first()
+        if server_block:
+            try:
+                extra = build_connection_block_extra(sync_org_task)
+                prefect_service.upsert_airbyte_connection_block(
```
himanshudube97, line 745: same issue here -> if pefect upsert fails then what ? we have already altered the db.

ddpui/ddpdbt/dbt_service.py
```diff
@@ -247,17 +247,6 @@ def delete_dbt_workspace(org: Org):
             pass
         secret_block.delete()
```
himanshudube97, line 248: When this function is called,  should it delete workspace.dbt_profile_secret_block too ? 
or not because now its a warehouse secret, not direclty linked only with dbt (but with the new cast op after sync too). 
just thinking.

Ishankoradia: nope, it shouldn't since its mapped to the warehouse & now airbyte syncs also need it

ddpui/core/pipelinefunctions.py
```diff
@@ -257,18 +288,17 @@ def pipeline_with_orgtasks(
...
     gitrepo_url: str = None,
 ):
     """
     Returns a list of task configs for a pipeline;
     This assumes the list of orgtasks is in the correct sequence
     """
     task_configs = []
-    # This block works perfectly for dbt cli tasks and dbt cloud tasks both.
+    warehouse = OrgWarehouse.objects.filter(org=org).first()
+    warehouse_secret_block = warehouse.dbt_profile_secret_block if warehouse else None
```
himanshudube97, line 301: Why this condition, we always need a warehouse right ? so if warehouse = None , throw error.

ddpui/ddpdbt/dbthelpers.py
```diff
@@ -205,13 +110,21 @@ def create_or_update_dbt_profile_secret_blk(
...
-    org.dbt.save(update_fields=["dbt_profile_secret_block"])
+
+    # FUTURE-TODO: remove the org.dbt assignment since we are moving it to warehouse.dbt_profile_secret_block
+    if org.dbt:
+        org.dbt.dbt_profile_secret_block = dbt_profile_secret_block_row
+        org.dbt.save(update_fields=["dbt_profile_secret_block"])
+
+    warehouse.dbt_profile_secret_block = dbt_profile_secret_block_row
+    warehouse.save(update_fields=["dbt_profile_secret_block"])
+
```
himanshudube97, line 121: Add all these inside a transaction

himanshudube97 (approved):

PR #1454, opened 2026-08-25
Add S3-backed image upload for dashboard text/image widgets

NaveenCode:
Summary

Add S3-backed image upload/delete endpoints for dashboard text/image widgets (PUT/DELETE /api/dashboards/images/), mirroring the existing org-logo upload pattern (same bucket, same s3_utils helpers, same 5MB limit).
New service functions upload_widget_image/delete_widget_image in dashboard_service.py: validate content type and file size, scope S3 keys to the org (orgs/{slug}/dashboards/images/...), and reject deletes of keys outside the caller's own org prefix.
New schemas WidgetImageUploadResponse/WidgetImageDeleteRequest in dashboard_schema.py.
Routes registered before the /{dashboard_id}/-pattern routes — fixes a bug where /images/ was otherwise matched as dashboard_id="images" by an earlier route.
Added 8 service-layer tests (validation, missing-bucket-env, S3 failure, success) and 4 API-layer tests 

## Summary by CodeRabbit

* **New Features**
  * Added support for uploading and deleting dashboard widget images.
  * Dashboard duplication and trial cloning now preserve widget images with organization-specific copies.
  * Added dashboard favorite and unfavorite actions; responses show favorite status and access levels.

* **Bug Fixes**
  * Improved access controls across dashboard, lock, and filter operations.
  * Dashboard deletion now removes associated sharing, access-request, and favorite records.

* **Changes**
  * Removed public dashboard sharing controls and status endpoints.


ddpui/api/dashboard_native_api.py
```diff
@@ -91,6 +100,62 @@ def list_dashboards(
...
+    the returned image_url inline in the widget's own config, persisted via
+    the normal dashboard save flow.
+    """
+    orguser: OrgUser = request.orguser
+    if orguser.org is None:
+        raise HttpError(400, "no associated org")
+
+    try:
+        image_url, image_key = upload_widget_image(
+            file_bytes=file.read(),
```
himanshudube97, line 134: we are directly loading the file in the memory. we should check the size before reading it and throw error if it exceeds our limit

ddpui/services/dashboard_service.py
```diff
@@ -1148,3 +1171,139 @@ def delete_dashboard_safely(dashboard_id: int, orguser: OrgUser) -> tuple[bool,
...
+                continue
+            try:
+                new_url, new_key = copy_widget_image(image_key, source_org, dest_org)
+            except WidgetImagePermissionError:
+                logger.warning(f"Rejecting out-of-org widget image reference {image_key}")
+                cfg["imageUrl"] = None
+                cfg["imageKey"] = None
+                continue
+            except WidgetImageStorageError as err:
+                logger.warning(f"Skipping widget image copy for {image_key}: {err.message}")
```
himanshudube97: so when this error occurs, the continue at line 1306 skips the loop and the cfg.imageUrl doesnt get new values and they still have original values only right ? so deep copy ina way failed ?

NaveenCode: Yes, it currently works this way. But if you'd prefer, we could make the duplicate fail entirely when the new image copy fails.I don't think we should do that just for an image-copy failure, since dashboard duplication involves many independent pieces — filters, tabs, multiple charts/KPIs, and possibly multiple images.

That said, we can still discuss this further, or show an error/warning on the duplicated dashboard instead of failing it outright. Let me know what you think.

cc: @pradnk

ddpui/api/dashboard_native_api.py
NaveenCode: updated, thanks

ddpui/api/dashboard_native_api.py
```diff
@@ -91,6 +101,67 @@ def list_dashboards(
...
+    except WidgetImageStorageError as err:
+        raise HttpError(502, err.message) from err
+
+    return WidgetImageUploadResponse(image_url=image_url, image_key=image_key)
+
+
+@dashboard_native_router.delete("/images/")
+@has_permission(["can_edit_dashboards"])
+def delete_dashboard_widget_image(request, payload: WidgetImageDeleteRequest):
+    """Delete a dashboard widget image from S3 (on explicit remove, or when
```
himanshudube97: we don't delete image from s3

ddpui/services/dashboard_service.py
```diff
@@ -1148,3 +1171,148 @@ def delete_dashboard_safely(dashboard_id: int, orguser: OrgUser) -> tuple[bool,
 
     logger.info(f"Dashboard '{dashboard_title}' deleted by {orguser.user.email}")
     return True, ""
+
+
+# =============================================================================
+# Widget Images (dashboard text/image widgets)
+# =============================================================================
+
+ALLOWED_WIDGET_IMAGE_CONTENT_TYPES = {"image/jpeg", "image/png", "image/gif", "image/webp"}
```
himanshudube97: move these to constant files. shouldn't be here in service file.

ddpui/utils/s3_utils.py
```diff
@@ -36,6 +36,28 @@ def upload_file(bucket: str, key: str, file_bytes: bytes, content_type: str) ->
...
+    """
+    region = _get_s3_region()
+    s3 = AWSClient.get_instance("s3")
+    s3.copy_object(Bucket=bucket, CopySource={"Bucket": bucket, "Key": source_key}, Key=dest_key)
+    url = _build_s3_url(bucket, region, dest_key)
+    logger.info(f"Copied file from s3://{bucket}/{source_key} to s3://{bucket}/{dest_key}")
+    return url
+
+
 def delete_file(bucket: str, key: str) -> None:
```
himanshudube97, line 61: do we need this function now ?

himanshudube97 (approved):

ddpui/utils/s3_utils.py
NaveenCode: Yes, we need this. This is used for the logo image; delete also

ddpui/api/dashboard_native_api.py
NaveenCode: yes we don't need this, it was a dead code , made the cleanup

ddpui/services/dashboard_service.py
NaveenCode: done,thanks

